PR #10, opened 2019-11-20
6 extract json

pchelle:


R/configuration-plan.R
```diff
@@ -0,0 +1,117 @@
...
+      print(paste(
+        "Reporting Engine: Initilization of Configuration Plan from",
+        jsonFile
+      ))
+      configurationPlan <- jsonlite::fromJSON(jsonFile)
+      self$reInputPath <- reInputPath
+      self$reOutputPath <- reOutputPath
+
+      if (dir.exists(reOutputPath)) {
+        warning(paste("Reporting Engine output path:", reOutputPath, "already exists"),
```
msevestre: @pchelle CAn we create a file like in TLF with all the constant? (messages.R or whatever we called it)? 
That way, if we need to correct a typo, we know where to look

R/configuration-plan.R
```diff
@@ -0,0 +1,117 @@
...
+      output = file.path(path, "_title.md")
+    )
+
+    # Copy Content
+    file.copy(
+      file.path(reInputPath, sectionsInput$Content[sectionIndex]),
+      file.path(path, "_content.md")
+    )
+
+    sectionsOutput[
```
msevestre: Can you explain what this section does?  I have no idea by looking at the code

R/configuration-plan.R
```diff
@@ -0,0 +1,117 @@
+#' @title ConfigurationPlan
+#' @docType class
+#' @description  Abstract class for ConfigurationPlan
```
msevestre: Not an abstract class

Also let's use this opportunity to document all parameter of the class right away. It's a pain to do during the dev but too much work to do after the fact

msevestre: see example of documentation of R6Class here
https://github.com/Open-Systems-Pharmacology/OSPSuite-R/blob/develop/R/quantity.R#L8

R/error-checks.R
```diff
@@ -0,0 +1,162 @@
...
+  if (nullAllowed && is.null(mapping)) {
+    return()
+  }
+
+  validateIsString(mapping)
+  variableNames <- names(data)
+
+  validateIsIncluded(mapping, variableNames)
+
+  return()
```
msevestre: last line return is not required

R/configuration-plan.R
msevestre: @pchelle Still needs explanation here. What is the intent of the code?

R/error-checks.R
```diff
@@ -0,0 +1,162 @@
...
+}
+
+#' Check if the provided object is of certain type
+#'
+#' @param object An object or a list of objects
+#' @param type String  representation or Class of the type that should be checked for
+#'
+#' @return TRUE if the object or all objects inside the list are of the given type.
+#' Only the first level of the given list is considered.
+isOfType <- function(object, type) {
```
msevestre: Let's copy all the tests from Core for now. Better to have it duplicate dif we duplicate the code as well

tests/testthat/test-multiplication.R
```diff
@@ -0,0 +1,3 @@
+test_that("multiplication works", {
```
msevestre: This needs to be removed and replaced with real tests

msevestre:
@pchelle I am going to merge here. Can you please make sure that for future PR, you implement the requested changes in a timely manner. Open PR without feedback always have a greater change to bring merge conflicts with them

PR #44, opened 2020-01-23
Mean model jan

abdullahhamadeh:


R/mean-model.R
```diff
@@ -0,0 +1,250 @@
...
+      #     addOutputs(quantitiesOrPaths = pathsPerCompoundList[[cmp]], simulation = loadedModel) #add outputs for current compound to loadedModel's output list
+      #   }
+      # }
+
+      #Simulate mean model (details s. #31)
+      #Remarks: Mass balance plots require all model outputs, thus it's not enough to save only model outputs of interest
+
+      # IF TIME RASTER PROVIDED THEN USE IT, ELSE USE DEFAULT SIMULATION TIMES IN PKML
+      self$simulationResults <- runSimulation(loadedModel)
+      self$simulationOutputs <- getOutputValuesTLF(simulationResults = self$simulationResults, quantitiesOrPaths = self$modelOutputsList , individualIds = 0 , population =  loadPopulation(csvPopulationFile = "./data/popData.csv"))  #using this placeholder population CSV file for now...need to make optional the population input to getOutputValuesTLF
```
msevestre: getOutputValuesTLF does not exist anymore in ospsuite.core. This needs to be updated

R/mean-model.R
```diff
@@ -0,0 +1,250 @@
...
+        }
+      }
+      self$massBalance <- as.data.frame(self$massBalance) #Convert massBalance list to datafram
+
+      # print(paste0(self$simulationOutputsFolder,"/",self$modelDisplayName,"-outputs.csv"))
+      if (saveMassBalance){
+        write.csv(self$massBalance,file = paste0(self$simulationOutputsFolder,"/",self$modelDisplayName,"-mass-balance.csv"))
+      }
+
+
```
msevestre: remove spaces

msevestre: @abdullahhamadeh Please run the `styler::tidyverse_style()` command (see install in R coding standards) on the whole package

R/mean-model.R
```diff
@@ -0,0 +1,250 @@
...
+      tempFileName <- paste0("./tests/tempPKresults_",as.numeric(Sys.time()),".csv")
+      exportPKAnalysesToCSV(pkAnalyses = self$PKAnalysis , filePath = tempFileName)
+      PKResultsDF <- read.csv(file = tempFileName, check.names = FALSE)
+      file.remove(tempFileName)
+      #print(PKResultsDF)
+      #print(PKResultsDF[["IndividualId"]])
+      #print(colnames(PKResultsDF))
+
+
+      userDefinedPKResults = list()
```
msevestre: Any reason why you are not using the construct <- list(IndividualId<-NULL, Parameter<-null etc..)

R/mean-model.R
```diff
@@ -0,0 +1,250 @@
...
+      userDefinedPKResults[["IndividualId"]]<-NULL
+      userDefinedPKResults[["Quantity Path"]]<-NULL
+      userDefinedPKResults[["Parameter"]]<-NULL
+      userDefinedPKResults[["Value"]]<-NULL
+      userDefinedPKResults[["Unit"]]<-NULL
+
+
+      for (pth in self$simulationResults$allQuantityPaths){
```
msevestre: let's use path instead of pth.

R/mean-model.R
```diff
@@ -0,0 +1,250 @@
...
+
+      ### SensitivityAnalysisRunOptions has two parameters
+      ### numberOfCoresToUse and showProgress.
+      ### These could be inputs to the function
+      ### runSensitivityAnalysis to simplify.
+
+
+      #next calculate pk parameter for the perturbed output vectors (which are functions of time)
+      for (output in loadedModel$outputSelections$allOutputs) {
+        pkSensitivities <- sensitivityAnalysisResults$allPKParameterSensitivitiesFor(pkParameterName = "AUC", outputPath = output$path)
```
msevestre: Hard coded string here? AUC? That should be a parameter no?

R/units-of-work.R
```diff
@@ -0,0 +1,289 @@
...
+#' @description Run a sensitivity analysis from a simulation
+#' @param simulation simulation class object
+#' @param pkParameterName name of parameter to be analyzed
+#' @param totalSensitivityThreshold numeric value between 0 and 1.
+#' Close to 0, only the most sensitive output paths are returned.
+#' Close to 1, almost all the output paths are returned.
+#' @return sensitivityResults
+#' @export
+#' @import ospsuite
+analyzeSensitivity <- function(simulation,
```
msevestre: let's create a file for each "type of unit of work" e.g. utilities-xxx
Otherwise this file is going to get out of control

R/utils.R
```diff
@@ -0,0 +1,127 @@
...
+#' @param sep character string separating path elements. "/" is default value.
+#' @return fileName character string of the trimmed filed name
+#' @description
+#' Trim path and extension of a file
+#' @examples
+#' \dontrun{
+#' pathName <- "folder/subfolder/testFile.txt"
+#' trimFileName(pathName, extension = "txt")
+#' }
+trimFileName <- function(path, extension = NULL, sep = "/") {
```
msevestre: Please write tests for all utilties function. We need to improve our test coverage here and this is a good place to start

R/units-of-work.R
abdullahhamadeh: Done

Issue #18, opened 2019-11-27
Population plots and tables: Time profiles and Residuals


Yuri05:
# Inputs
  * Output folder of the _Simulate population_ task (with calculated simulation results)
  * Output Folder for results (all results will be generated here)
  * OPTIONAL Number of Cores
  * Type. One of:
      * **Pediatric**: one population _may be_ marked as (adult) reference
      * **Parallel comparison**: no population is marked as special.
      * **Ratio comparison**:
        * special case for analysis of one property of a population (e.g. GFR)
        * one population (e.g. "healthy") is marked as reference
        * for all other (e.g. "diseased") populations: ratios of all PK Parameters to the corresponding healthy one is calculated and plotted as Box-Whisker plot
  * 1 or more populations
  * List of datasets
  * Settings (s. also #13):
      * Reference population ( may be empty)
        * OPTIONAL Time offset
      * display Unit for time. **Default: h**
      * List of time Time ranges.  One plot per time range
         * Single dose: Default total range
         * For multiple dose additionally: first and last application 
      * scale of y axes (lin/log/both). **Default: both**
      * Aggregation type for Residuals. One of {_Mean_, _Geomean_, _Median_}. Default: Median 
      * Types of Plot. **Default: all 6 plot types**. For each plot: plot per pop or for all pops or both
        1. Simulated time profile: Mean, Median and 5th - 95th percentile range of the simulated time profiles (if available with data). If a reference population is defined, it is also plotted.
        2. Residuals (based on population mean or geomean or median (depending on the aggregation type) ) vs time
        3. Residuals (based on population mean or geomean or median (depending on the aggregation type) ) vs simulated output
        4. Predicted (population mean or geomean or median (depending on the aggregation type)) vs observed
        5. Histogram of residuals (**For all outputs together**)
        6. qq-plot of residuals (**For all outputs together**)

# Outputs
In the output folder: all plots/tables as specified in the settings
* one subsection for each population, reference population is plotted in each population as reference
* **(for all plot types except for time profiles)** if specified in the settings: one section for merged plots (= all pops in one plot)
[...]

PR #53, opened 2020-02-11
39 mean model jan 27 develop

pchelle:
I branched from Abdullah's pull request (so I closed both our pull request since this one is the most up-to-date).
The following code includes Abdullah's pull request with documentation updates fixing some of the R6 warnings (there are still a couple of R6 classes I need to fix).
I also drafted default nomenclature for files and folder names + logging functions that will need to be included within the tasks.
I also added Juri's first example in the dev folder.

There was one missing piece to our workflow that Juri's examples highlighted: an R framework for the matlab variable OutputList.
So, I drafted two R6 classes for MeanModelSet and PopulationSet that aims at providing this R framework.
Similarly, PopulationWorkflow was the R framework for PopRunSet and 
MeanModelWorkflow was the R framework for MeanModelSet

Let me know if it is okay to merge this pull request and move forward from here.


Yuri05:
@pchelle please remove _.RHistory*_ files from PR

Yuri05:
@pchelle @abdullahhamadeh Question: why did you add the matlab example here? 
(under `tests/dev/01_MeanModel_Raltegravir`)
When later on we create a corresponding R example, I would store in this repo only R-inputs/outputs but not the Matlab stuff. Or do I miss something?

pchelle:
yes, indeed.
I am going to keep only the pkml and the data then.

R/calculate-pk-parameters-task.R
```diff
@@ -0,0 +1,37 @@
+#' @title CalculatePKParametersTask
+#' @docType class
+#' @description  Calculating PK parameters task settings for Reporting Engine
+#' @field resultsFileName string with no file extension.  Name of CSV file to which simulation results will be saved.
+#' @section Methods:
```
msevestre: @pchelle @abdullahhamadeh 
For new classes, please update to new R6 doc conventions
e.g. you can document methods right away where the method is defined

See ospsuite-r for example
https://github.com/Open-Systems-Pharmacology/OSPSuite-R/blob/develop/R/population.R

R/population-workflow.R
```diff
@@ -26,251 +16,441 @@ PopulationWorkflow <- R6::R6Class(
...
+    #' Define plot goodness of fit `task` settings
+    #' @param input file or folder of input
+    #' @param output file or folder of output
+    #' @param settings specific settings for task
+    #' @param active logical indicating if `task` is performed in worklfow.
+    #' Default value is `TRUE`
+    #' @param message title of the `task`.
+    #' Default value indicates `task` name.
+    #' @return A new `Task` object
+    plotGoFSettings = function(input = NULL,
```
Yuri05: This corresponds to "Time profiles and residuals" task, correct? https://github.com/Open-Systems-Pharmacology/OSPSuite.ReportingEngine/issues/18
Then it might be better to call it `TimeProfilesAndResiduals` instead of `GOF`.

R/utilities-logging.R
```diff
@@ -0,0 +1,56 @@
+#' @title logError
+#' @description Save error messages into a log error file
+#' @param message message to save in log file
+#' @param file log file name.
+#' Default is `log-error.txt`
+#' @param printConsole logical to print error on console
```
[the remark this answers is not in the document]
pchelle: The goal of this parameter is to have a way of printing the error or information messages in the console while it is running on the user R session.
If FALSE, it is silent and only printed within the log.
I was thinking that such parameter could be useful for logDebug since you do not really want to have every intermediate printing on the console.
If you think it is not needed, I can remove it. Otherwise, I can define instead a silent mode  parameter which should be clearer.

abdullahhamadeh:
Should be good to merge now.

PR #55, opened 2020-02-18
Extend pop sens analysis index file

abdullahhamadeh:
A population sensitivity analysis consists of potentially many sensitivity analysis evaluated for models of different individuals in the population.

Each individual will have their own SA results CSV files.

Along with these result CSV files, there will be output an index CSV file that lists the sensitivity analyses conducted.

Each entry in the index file will list the individuals that yield pk parameter values at user-specified quantiles of the population-wide distribution of the pk parameter.  The entry will also give the value and unit of the pk parameter at each quantile.  The last field of the entry will be the name of the CSV file with that individual's sensitivity analysis.


R/utilities-sensitivity-analysis.R
```diff
@@ -1,342 +1,343 @@
...
+#' @param resultsFileName root name of population sensitivity analysis results CSV files
+#' @return SA results for individual or population
+#' @export
+#' @import ospsuite
+analyzeSensitivity <- function(simFilePath,
+                               parametersToPerturb = NULL,
+                               popFilePath = NULL,
+                               individualID = NULL,
+                               numberOfCores = 1,
+                               resultsFileFolder = "getwd()",
```
msevestre: Not sure about this default here. This should probably be always set

R/utilities-sensitivity-analysis.R
```diff
@@ -1,342 +1,343 @@
...
+#' @param parametersToPerturb paths to parameters to perturb in sensitivity analysis
+#' @param popFilePath path to the population data file
+#' @param individualID ID of individual in population data file for whom to perform sensitivity analysis
+#' @param numberOfCores number of cores over which to parallelize the sensitivity analysis
+#' @param resultsFileFolder path to population sensitivity analysis results CSV files
+#' @param resultsFileName root name of population sensitivity analysis results CSV files
+#' @return SA results for individual or population
+#' @export
+#' @import ospsuite
+analyzeSensitivity <- function(simFilePath,
```
msevestre: I am bit confused with the terminology
We are dealing with `SensitivityAnalysis`
Here you analyze the sentisivity analysis. But in fact you are running the sensitivity analysis.? 
Shouldn't we call it runSensitivity and that would be the only exported method? All ther other one private?

R/utilities-sensitivity-analysis.R
```diff
@@ -1,342 +1,343 @@
...
-getIndividualSAResultsFileName <- function(individualId,resultsFileName){
-  return(paste(resultsFileName,"IndividualId",individualId,sep="-"))
-}
+#' @title analyzeSensitivity
+#' @description Determine whether to run SA for individual or population.  If for individual,  pass simulation to individualSensitivityAnalysis.
+#' If SA is for population, loop thru population file, extract parameters for each individual, and pass them to individualSensitivityAnalysis.
+#' @param simFilePath path to simulation file
+#' @param parametersToPerturb paths to parameters to perturb in sensitivity analysis
+#' @param popFilePath path to the population data file
+#' @param individualID ID of individual in population data file for whom to perform sensitivity analysis
```
msevestre: This looks like this is only one individual. But this can be an array right? Maybe it should be individualIDs?
Also what do we use as nomenclature in core ?ID or Id? Let's make sure we are consistent (I can't remember)

R/utilities-sensitivity-analysis.R
```diff
@@ -1,342 +1,343 @@
...
+#' @description Run SA for an individual, possibly after modifying the simulation using individualParameters.  Determine whether to run SA for on single core or in parallel.
+#' If on single core, pass simulation to analyzeCoreSensitivity.  If in parallel, pass simulation to runParallelSensitivityAnalysis.
+#' @param simFilePath path to simulation file
+#' @param parametersToPerturb paths to parameters to perturb in sensitivity analysis
+#' @param individualParameters is an object storing an individual's parameters, obtained from a population object's getParameterValuesForIndividual() function.
+#' @param numberOfCores is the number of cores over which to parallelize the sensitivity analysis
+#' @param resultsFileFolder path to population sensitivity analysis results CSV files
+#' @param resultsFileName root name of population sensitivity analysis results CSV files
+#' @return allResultsFileNames, the paths to CSV files containing results of sensitivity analysis
+#' @export
```
msevestre: Why does it need to be exported?

R/utilities-sensitivity-analysis.R
```diff
@@ -1,342 +1,343 @@
...
+      resultsFileName
+    )
+  } else {
+    # No parallelization
+    updateSimulationIndividualParameters(simulation = sim, individualParameters)
+    allResultsFileNames <- file.path(resultsFileFolder, paste0(resultsFileName, ".csv"))
+    analyzeCoreSensitivity(
+      simulation = sim,
+      parametersToPerturb = parametersToPerturb,
+      totalSensitivityThreshold = 1,
```
msevestre: That should be  a parameter

R/utilities-sensitivity-analysis.R
```diff
@@ -1,342 +1,343 @@
...
+
+  # Update simulation with individual parameters
+  Rmpi::mpi.bcast.cmd(updateSimulationIndividualParameters(simulation = sim, individualParameters))
+
+
+
+  Rmpi::mpi.remote.exec(analyzeCoreSensitivity(
+    simulation = sim,
+    parametersToPerturb = listSplitParameters[[mpi.comm.rank()]],
+    totalSensitivityThreshold = 1,
```
msevestre: same here. Parameter

R/utilities-sensitivity-analysis.R
```diff
@@ -1,342 +1,343 @@
...
+
+  Rmpi::mpi.bcast.Robj2slave(obj = allResultsFileNames)
+
+  # Load simulation on each core
+  Rmpi::mpi.bcast.cmd(sim <- loadSimulation(simFilePath))
+
+  # Update simulation with individual parameters
+  Rmpi::mpi.bcast.cmd(updateSimulationIndividualParameters(simulation = sim, individualParameters))
+
+
```
msevestre: Remove spaces

R/utilities-sensitivity-analysis.R
```diff
@@ -1,342 +1,343 @@
...
+  Rmpi::mpi.bcast.cmd(library("ospsuite.reportingengine"))
+  Rmpi::mpi.bcast.Robj2slave(obj = simFilePath)
+  Rmpi::mpi.bcast.Robj2slave(obj = listSplitParameters)
+  Rmpi::mpi.bcast.Robj2slave(obj = resultsFileFolder)
+  Rmpi::mpi.bcast.Robj2slave(obj = individualParameters)
+
+  # Generate a listcontaining names of SA CSV result files that will be output by each core
+  allResultsFileNames <- generateResultFileNames(numberOfCores = numberOfCores, folderName = resultsFileFolder, fileName = resultsFileName)
+
+
```
msevestre: remove spaces

R/utilities-sensitivity-analysis.R
```diff
@@ -1,342 +1,343 @@
...
+  Rmpi::mpi.bcast.cmd(updateSimulationIndividualParameters(simulation = sim, individualParameters))
+
+
+
+  Rmpi::mpi.remote.exec(analyzeCoreSensitivity(
+    simulation = sim,
+    parametersToPerturb = listSplitParameters[[mpi.comm.rank()]],
+    totalSensitivityThreshold = 1,
+    resultsFilePath = allResultsFileNames[mpi.comm.rank()],
+    numberOfCoresToUse = 1 # Number of local cores, set to 1 when parallelizing.
```
msevestre: YES!

R/utilities-sensitivity-analysis.R
```diff
@@ -1,342 +1,343 @@
...
+#' @param simulation simulation class object
+#' @param parametersToPerturb paths of parameters to be analyzed
+#' @param totalSensitivityThreshold numeric value between 0 and 1.
+#' Close to 0, only the most sensitive output paths are returned.
+#' Close to 1, almost all the output paths are returned.
+#' @param resultsFilePath Path to file storing results of sensitivity analysis
+#' @param numberOfCoresToUse Number of cores to use on local node.  This parameter should be should be set to 1 when parallelizing over many nodes.
+#' @return Save sensitivity analysis results as CSV in path given by resultsFilePath.
+#' @import ospsuite
+#' @export
```
msevestre: Do we need to export it?

R/utilities-sensitivity-analysis.R
```diff
@@ -1,342 +1,343 @@
...
+analyzeCoreSensitivity <- function(simulation,
+                                   parametersToPerturb = NULL,
+                                   totalSensitivityThreshold = 1,
+                                   resultsFilePath = paste0(getwd(), "sensitivityAnalysisResults.csv"),
```
msevestre: not sure about this default path here.

R/utilities-sensitivity-analysis.R
```diff
@@ -1,342 +1,343 @@
...
+
+
+  if (is.null(numberOfCoresToUse)) {
+    sensitivityAnalysisRunOptions <- SensitivityAnalysisRunOptions$new(showProgress = FALSE)
+  }
+  else {
+    sensitivityAnalysisRunOptions <- SensitivityAnalysisRunOptions$new(
+      showProgress = FALSE,
+      numberOfCoresToUse = numberOfCoresToUse
+      #The numberOfCoresToUse input in the SensitivityAnalysisRunOptions initializer should not be set to NULL.
```
msevestre: Interesting. That looks like a bug to me don't you think?

R/utilities-sensitivity-analysis.R
```diff
@@ -1,342 +1,343 @@
...
+    )
+  }
+  print("Running sensitivity analysis...")
+  sensitivityAnalysisResults <- runSensitivityAnalysis(
+    sensitivityAnalysis = sensitivityAnalysis,
+    sensitivityAnalysisRunOptions = sensitivityAnalysisRunOptions
+  )
+  print("...done")
```
msevestre: use logger

R/utilities-sensitivity-analysis.R
```diff
@@ -1,342 +1,343 @@
...
+    sensitivityAnalysisRunOptions <- SensitivityAnalysisRunOptions$new(showProgress = FALSE)
+  }
+  else {
+    sensitivityAnalysisRunOptions <- SensitivityAnalysisRunOptions$new(
+      showProgress = FALSE,
+      numberOfCoresToUse = numberOfCoresToUse
+      #The numberOfCoresToUse input in the SensitivityAnalysisRunOptions initializer should not be set to NULL.
+    )
+  }
+  print("Running sensitivity analysis...")
```
msevestre: use logger

R/utilities-sensitivity-analysis.R
```diff
@@ -1,342 +1,343 @@
...
+}
+
+
+
+
+#' @title getQuantileIndividualIds
+#' @description Find IDs of individuals whose PK analysis results closest to quantiles given by vector of quantiles quantileVec
+#' @param pkAnalysisResultsDataframe Dataframe storing the PK analysis results for multiple individuals for a single PK parameter and single output path
+#' @return ids, IDs of individuals whose PK analysis results closest to quantiles given by vector of quantiles quantileVec
+getQuantileIndividualIds <- function(pkAnalysisResultsDataframe, quantileVec = c(0.05, 0.5, 0.95)) {
```
msevestre: Default quantile should be part of the method descriptions

R/utilities-sensitivity-analysis.R
```diff
@@ -1,342 +1,343 @@
...
+#' The individuals yielding pk parameters closest to these quantiles will be selected for sensitivity analysis.  Default is (0.05,0.5,0.95)
+#' @param numberOfCores the number of cores to be used for parallelization of the sensitivity analysis.  Default is 1 core (no parallelization).
+#' @export
+populationSensitivityAnalysis <- function(simFilePath,
+                                          popDataFilePath,
+                                          pkParameterResultsFilePath,
+                                          resultsFileFolder,
+                                          resultsFileName,
+                                          popSAResultsIndexFile = "sensitivityAnalysesResultsIndexFile",
+                                          quantileVec = c(0.05, 0.5, 0.95),
```
msevestre: We duplicate the same default values here. This should be in only one method and the other one should require inputs in that case

R/utilities-sensitivity-analysis.R
```diff
@@ -1,342 +1,343 @@
...
+  individualIdColumn <- NULL
+  valuesColumn <- NULL
+  unitsColumn <- NULL
+  quantileColumn <- NULL
+
+  for (output in outputs) {
+    for (pkParameter in pkParameters) {
+      singleOuputSinglePKDataframe <- allPKResultsDataframe[ allPKResultsDataframe["QuantityPath"] == output & allPKResultsDataframe["Parameter"] == pkParameter, ]
+      quantileResults <- getQuantileIndividualIds(singleOuputSinglePKDataframe, quantileVec)
+      for (n in 1:length(quantileResults$ids)) {
```
msevestre: what about we use `i` instead of` n`?  Or has n a special meaning?

R/utilities-sensitivity-analysis.R
```diff
@@ -1,342 +1,343 @@
...
+#' @param resultsFileName root name of population sensitivity analysis results CSV files
+#' @param popSAResultsIndexFile name of CSV file that will store index that identifies name of sensitivity analysis results file for each sensitivity analysis run
+
+getSAFileIndex <- function(pkParameterResultsFilePath,
+                           quantileVec,
+                           resultsFileFolder,
+                           resultsFileName,
+                           popSAResultsIndexFile) {
+
+  allPKResultsDataframe <- getPKResultsDataFrame(pkParameterResultsFilePath)
```
msevestre: I am not sure once again but I am wondering if we need the data frame at all. We have an object (PKParameterResults) or something like that that can return all values for a given output, pk parameter etc... I feel that we are constructing a data frame based on the PKResults just to get the same behavior. Of course, I may miss the point completely but I just wanted to raise the point :)

R/utilities-sensitivity-analysis.R
abdullahhamadeh: Done

R/utilities-sensitivity-analysis.R
abdullahhamadeh: Renamed it to runSensitivity

populationSensitivityAnalysis is another function in this file that needs to be exported because it is accessed by the workflow class.

I had kept a few other methods exported in case user needs quick & more direct access for very specific functions.  eg 
1 - individualSensitivityAnalysis is used to perform SA for one individual (ie for a specific individual ID, with individual parameters coming from a population ), possibly in parallel
2 - analyzeCoreSensitivity is used to perform SA for an given simulation on one core.  Mainly used for the parallel runs.  

removing export for these latter two for now

R/utilities-sensitivity-analysis.R
abdullahhamadeh: It is only for one individual.  runSensitivity loops through individual id arrays as need and passes the id number to this function.  

renamed to individualId

R/utilities-sensitivity-analysis.R
abdullahhamadeh: removed

R/utilities-sensitivity-analysis.R
abdullahhamadeh: removing totalSensitivityThreshold for now, adding variationRange instead

R/utilities-sensitivity-analysis.R
abdullahhamadeh: actually we need analyzeCoreSensitivity exported for access by cores

R/utilities-sensitivity-analysis.R
abdullahhamadeh: removed the default here

R/utilities-sensitivity-analysis.R
abdullahhamadeh: done

R/utilities-sensitivity-analysis.R
abdullahhamadeh: default is now only specified at the workflow level, rather than in the function

R/utilities-sensitivity-analysis.R
abdullahhamadeh: Yes this one should be exported for access by cores when parallelizing

R/utilities-sensitivity-analysis.R
abdullahhamadeh: Seems so

R/utilities-sensitivity-analysis.R
abdullahhamadeh: done

R/utilities-sensitivity-analysis.R
abdullahhamadeh: done

R/utilities-sensitivity-analysis.R
abdullahhamadeh: done

R/utilities-sensitivity-analysis.R
abdullahhamadeh: done

R/utilities-sensitivity-analysis.R
abdullahhamadeh: The dataframe format was convenient for the later part of this function

PR #61, opened 2020-02-26
Run sa for only some parameters

abdullahhamadeh:


R/sensitivity-analysis-task.R
```diff
@@ -10,23 +10,26 @@ SensitivityAnalysisTask <- R6::R6Class(
...
 
     #' @description
     #' Create a `SensitivityAnalysisTask` object
     #' @param variationRange variation range for sensitivity analysis
     #' @param numberOfCores number of cores for parallel computation
-    #' @param generatedResultFileNames name of files where PK parameters are saved
+    #' @param parametersToPerturb parameters to perturb when performing sensitivity analysis
     #' @param ... parameters inherited from R6 class `Task` object
     #' @return A new `SensitivityAnalysisTask` object
     initialize = function(
```
msevestre: It would be good to add validations here as we do in core and tlf to ensure that parameters hvea the right type
Also `parametersToPerturb `  Are those instances of the `PArameter` class in core?

R/utilities-sensitivity-analysis.R
```diff
@@ -77,11 +77,11 @@ individualSensitivityAnalysis <- function(simFilePath,
...
   sim <- loadSimulation(simFilePath)
-  if (is.null(parametersToPerturb)) {   # If no perturbation parameters specified, perturb all parameters
+  if (is.null(parametersToPerturb)) { # If no perturbation parameters specified, perturb all parameters
     parametersToPerturb <- ospsuite::potentialVariableParameterPathsFor(simulation = sim)
   }
   totalNumberParameters <- length(parametersToPerturb)
-  numberOfCores <- min(numberOfCores, totalNumberParameters)   # In case there are more cores specified in numberOfCores than there are parameters, ensure at least one parameter per spawned core
+  numberOfCores <- min(numberOfCores, totalNumberParameters) # In case there are more cores specified in numberOfCores than there are parameters, ensure at least one parameter per spawned core
   if (totalNumberParameters == 0) {
     stop("No variable parameters found for sensitivity analysis.")
```
msevestre: Let's also log this as error in the log

R/sensitivity-analysis-task.R
abdullahhamadeh: `parametersToPerturb` is of the same format as the output of `ospsuite::potentialVariableParameterPathsFor()`, ie a list of strings of paths.  The two are handled the same way.

msevestre: Ok shouldn't we rename it to avoid confusion? At least the doc on the parameter should clarify this

abdullahhamadeh: to `variableParameterPaths` ok?

Issue #27, opened 2019-11-30
Add all error message in a message file instead of having strings defined everywhere in the code


msevestre:
Follow up of PR #10

PR #64, opened 2020-02-27
Add sa error check

abdullahhamadeh:


R/utilities-sensitivity-analysis.R
```diff
@@ -77,8 +77,23 @@ individualSensitivityAnalysis <- function(simFilePath,
...
+
+
   if (is.null(variableParameterPaths)) { # If no parameters to vary specified, vary all parameters valid for sensitivity analysis
-    variableParameterPaths <- ospsuite::potentialVariableParameterPathsFor(simulation = sim)
+    variableParameterPaths <- allVariableParameterPaths
+  } else {
+    #if a variableParameterPaths input is provided, ensure that all its elements exist within allVariableParameterPaths.  If not, give an error.
+
+    if(!all( variableParameterPaths %in% allVariableParameterPaths) ){
+      msg<- paste("Variable parameter path(s)",
```
msevestre: Let's define a single file with all the error messages as we have done with other libs. That way, if there is a typo or the message needs to change, we only look at the one place

msevestre: see issue https://github.com/Open-Systems-Pharmacology/OSPSuite.ReportingEngine/issues/27

R/utilities-sensitivity-analysis.R
```diff
@@ -77,8 +77,23 @@ individualSensitivityAnalysis <- function(simFilePath,
...
+  allVariableParameterPaths <- ospsuite::potentialVariableParameterPathsFor(simulation = sim)
+
+
   if (is.null(variableParameterPaths)) { # If no parameters to vary specified, vary all parameters valid for sensitivity analysis
-    variableParameterPaths <- ospsuite::potentialVariableParameterPathsFor(simulation = sim)
+    variableParameterPaths <- allVariableParameterPaths
+  } else {
+    #if a variableParameterPaths input is provided, ensure that all its elements exist within allVariableParameterPaths.  If not, give an error.
+
+    if(!all( variableParameterPaths %in% allVariableParameterPaths) ){
```
msevestre: This would fail if someone tries to vary a formula parmaeter for example that is not in the list of potentialxxx. But Ithink this is ok in that case. Thoughts @Yuri05 ?

R/utilities-sensitivity-analysis.R
abdullahhamadeh: Changed. I used validateIsIncluded.   Should we add logError commands to all error checks? @pchelle

msevestre: Would it make sense to defien your own stop method that basiaclly logError and then stops?

abdullahhamadeh: Added this to error-checks.R

PR #66, opened 2020-03-02
39 mean model workflow example

pchelle:
Here is the pull request for the mean model workflow example 01
Sorry, I have left many TO DOs in the files...
I haven't been able to get the last functions of the OSP Suite R so far, so I could not use the latest toUnit functions. Hopefully, I will make it work soon.


R/task.R
```diff
@@ -70,3 +70,31 @@ Task <- R6::R6Class(
...
   )
 )
+
+#' @title activateWorkflowTasks
+#' @description activates a series of `Tasks` from a `Workflow`
+#' @param workflow `MeanModelWorklfow` or `PopulationWorklfow` object
+#' @param tasks names of the tasks to activate
+#' Default activates all tasks of the workflow using workflow method `workflow$getAllTasks()`
+#' @export
+activateWorkflowTasks <- function(workflow, tasks = workflow$getAllTasks()) {
```
msevestre: Should not that be a method of the workflow ?

R/utilities-mass-balance.R
```diff
@@ -38,3 +38,342 @@ envList2PathStringsVector <- function(envList) {
...
+#' @return list of `ggplot` objects
+#' @export
+#' @import tlf
+#' @import ospsuite
+#' @import utils
+plotMeanMassBalance <- function(structureSet,
+                                plotConfigurations = NULL,
+                                selectedCompoundNames = NULL) {
+  simulation <- ospsuite::loadSimulation(structureSet$simulationSet$simulationFile)
+  molecules <- ospsuite::getAllMoleculesMatching(paths = "**", container = simulation)
```
msevestre: WE have a new method here that is going to be at least a factor 100 faster in ospsuite. 
getAllMoleculePaths or somethings`

R/utilities-mass-balance.R
```diff
@@ -38,3 +38,342 @@ envList2PathStringsVector <- function(envList) {
...
+    y = "Amount",
+    color = "Legend"
+  )
+
+  # TO DO: remove Testing Data Watermark when plot is complete
+  plotConfiguration <- plotConfiguration %||% tlf::PlotConfiguration$new(
+    data = data,
+    metaData = metaData,
+    dataMapping = timeVsAmountDataMapping,
+    watermark = "Testing !"
```
msevestre: Should not the watermark be part of the config of this workflow run? E.g. a settings that says this is STILL not QCed or something?

R/workflow.R
```diff
@@ -15,67 +14,105 @@ Workflow <- R6::R6Class(
...
+
+      taskNames <- names(isTaskVector[as.logical(isTaskVector)])
+
+      return(taskNames)
+    },
+
+    #' @description
+    #' Get a vector with all the names of active tasks within the `Workflow`
+    #' @return Vector of active `Task` names
+    getActiveTasks = function() {
```
msevestre: maybe a private method that can be called by getActiveTasks and getInactiveTasks
The code is exactly the same. 
getTasksWithStatus(bool status) and then you have in your code 

if (self[[taskName]]$active==status) {

R/task.R
pchelle: Yes, 
in case, I kept the methods within and outside of the workflow, so the user can choose both ways.

R/utilities-mass-balance.R
pchelle: Yes, for this I am going to define a default theme for the Reporting Engine.

R/workflow.R
pchelle: Great idea for preventing duplication of code
By the way, I feel I am currently duplicating/triplicating quite a few chunks of code for the plot tasks of the mean model workflow:
The steps seem kind of generic :
1) Initialization of the task
- Logging and writing the section title of the report
- Check that the input is available 
- Create an output folder to save the results
2) Run the task
- Logging
- Call the right functions with the right options (this is where they mainly differ) 
3) Termination of the task
- Logging
- Saving the output plots/results of the task
- Writing the report chunks

It may be worth creating a subclass "PlotTask".
So within the Workflow run, I can loop through all tasks with this subclass, to prevent re-using 4 times similar chunks of code.

PR #69, opened 2020-03-12
Parallel mode error logging

abdullahhamadeh:


R/utilities-calculate-pk-parameters.R
```diff
@@ -39,10 +39,11 @@ UserDefinedPKFunction <- R6::R6Class(
...
 )
 
-#' @title AllPKParameters
-#' @description  list of usual PK parameters from OSP Suite
+
+#' @title AllPKParameterNames
+#' @description  array of standard PK parameters from OSP Suite
 #' @export
-AllPKParameters <- enum(c(
+AllPKParameterNames <- c(
```
msevestre: This duplicates the enum that we have already in core. Why ?

abdullahhamadeh: What is it called in ospsuite?  I will try to use add it instead of this

msevestre: StandardPKParameter I believe

I am working on updating OSPSuite-R to support the global definition of user defined pk. Should be done by Monday

abdullahhamadeh: ospsuite::StandardPKParameter won't work here.  We need an array of the pk parameter names.  But I agree the list should be common to both packages.  Ok to merge this for now?  I think there are more changes to come soon...

msevestre: Why do you need to know the name explicitly? I am going to merge but something is still not clear to me.

abdullahhamadeh: I have added the option to specify which pk parameters the population SA is calculated for.  The user enters their choice as, eg, c("C_max","AUC").  I was using validateIsIncluded to check that these entries are in fact in the list of pk parameters that would be found in a pk calculation result file.  

I can use names(ospsuite::StandardPKParameter) for this check, but this returns pk parameter names that are different to the ones in a pk calculation results file.  For example, in names(ospsuite::StandardPKParameter) there is "Cmax" and in the results file there is "C_max".

msevestre: >I have added the option to specify which pk parameters the population SA is calculated for.

The PK Parameters are always calculated for ALL Parameters. Are you talking about which parameters should be displayed?

msevestre: I am going to create an issue to review that once the changes in PK Parameters are implemented. Since the user will also want to use user defined parameters, I don't think we can validate the name as it might be unknown.

abdullahhamadeh: For a single individual yes.  But it the population SA it worked like this before, for a given variation parameter p
1 - specify a list of quantiles eg (c(0.25,0.5,0.75))
2 - for each pk parameter the population's distribution of that pk parameter is analyzed to find individuals in the distribution closest to the quantiles
3 - for those individuals only, the sensitivity is calculated for ALL pk parameters (as you said) in response to variations in parameter p.

What I did is I added the option to only do step 2 for a selection of pk parameters.  Step 3 is unchanged, ie for a given individual, sensitivity of all pk parameters with respect to p is calculated.

msevestre: Oh I see. 
At any rate, with predefined ok parameters and user defined pk parameters, we have to come up with a good way of checking this

PR #82, opened 2020-03-16
Debug pop workflow example 3

abdullahhamadeh:


R/utilities-sensitivity-analysis.R
```diff
@@ -24,6 +24,34 @@ runSensitivity <- function(simFilePath,
   # If there is a population file and no individualId then do SA for entire population
   # If there is no population file and individualId then do SA for mean model
   # If there is no population file and no individualId then do SA for mean model.
+
+  sim <- loadSimulation(simFilePath)
+
+  allVariableParameterPaths <- ospsuite::potentialVariableParameterPathsFor(simulation = sim)
+
+  if (is.null(variableParameterPaths)) { # If no parameters to vary specified, vary all parameters valid for sensitivity analysis
```
msevestre: Can you please put the comment above the text it's commenting not besides

msevestre: Also don't hesitate to put comment on multiple lines so that we do not have to scroll to read them

The goal is to avoid scrolling left to right as it disturbs the flow. Think 120 col ish max

R/utilities-sensitivity-analysis.R
```diff
@@ -24,6 +24,34 @@ runSensitivity <- function(simFilePath,
...
+    variableParameterPaths <- allVariableParameterPaths
+  } else {
+    # if a variableParameterPaths input is provided, ensure that all its elements exist within allVariableParameterPaths.  If not, give an error.
+    validateIsIncluded(variableParameterPaths, allVariableParameterPaths)
+  }
+  totalNumberParameters <- length(variableParameterPaths)
+  numberOfCores <- min(numberOfCores, totalNumberParameters) # In case there are more cores specified in numberOfCores than there are parameters, ensure at least one parameter per spawned core
+  if (totalNumberParameters == 0) {
+    msg <- "No variable parameters found for sensitivity analysis."
```
msevestre: define all error messages in the message file

R/utilities-sensitivity-analysis.R
```diff
@@ -24,6 +24,34 @@ runSensitivity <- function(simFilePath,
...
+    logError(msg)
```
msevestre: don't we have logAndStop or sthg like this?

R/utilities-sensitivity-analysis.R
```diff
@@ -24,6 +24,34 @@ runSensitivity <- function(simFilePath,
...
+    stop(msg)
+  }
+
+  if (numberOfCores > 1){
```
msevestre: what if numberOfCores ==1? Is it run on the master node?

PR #90, opened 2020-03-17
Remove update individual parameter loop

abdullahhamadeh:


R/utilities-simulation.R
```diff
@@ -107,19 +107,11 @@ numberOfCores <- min(numberOfCores,numberOfIndividuals)
...
-          simulation = sim
-        )
-      },
-      simulation,
-      individualParameters
-    )
+  if (is.null(individualParameters)){
+    return()
+  }
+  else {
```
msevestre: no need for else here

PR #91, opened 2020-03-18
Change bcast.cmd to remote.exec

abdullahhamadeh:


R/utilities-sensitivity-analysis.R
```diff
@@ -170,36 +169,32 @@ runParallelSensitivityAnalysis <- function(simFilePath,
...
-
   tempLogFileNamePrefix <- file.path(defaultFileNames$workflowFolderPath(), "logDebug-core-sensitivity-analysis")
   tempLogFileNames <- paste0(tempLogFileNamePrefix, seq(1, numberOfCores))
+
+  # Generate a listcontaining names of SA CSV result files that will be output by each core
+  allResultsFileNames <- generateResultFileNames(numberOfCores = numberOfCores, folderName = resultsFileFolder, fileName = resultsFileName)
   logDebug(message = "Starting sending of parameters to cores")
   Rmpi::mpi.bcast.Robj2slave(obj = listSplitParameters)
   Rmpi::mpi.bcast.Robj2slave(obj = tempLogFileNamePrefix)
   Rmpi::mpi.bcast.Robj2slave(obj = individualParameters)
```
msevestre: Can you explain what this line does again when you have time?

abdullahhamadeh: Which line number? 191?

msevestre: yes. In general. Robj2slave

abdullahhamadeh: The cores don't have direct access to variables on master.  So `Robj2slave` is used to send R objects to the cores.  

`Rmpi::mpi.bcast.Robj2slave(obj = individualParameters)` is sending the parameters specific to the individual so that the simulation can be updated ahead of the SA.  

`Rmpi::mpi.bcast.Robj2slave(obj = listSplitParameters)` is sending the list of variation parameters that will be perturbed in the SA.

msevestre: sure. But what does that mean. Does that mean that a slave will have access to a variable called individualParameters?

I am assuming this only works with primitive types (string, boolean, numbers and list thereof e)?

abdullahhamadeh: Yes to both :)

msevestre: ok cool. Thanks :)

PR #104, opened 2020-03-27
Update population output folder structure

abdullahhamadeh:


R/population-sensitivity-analysis-settings.R
```diff
@@ -0,0 +1,103 @@
+#' @title PopulationSensitivityAnalysisSettings
+#' @description  R6 class for Population Sensitivity Analysis Settings
+#' @field variationRange variation range for sensitivity analysis
+#' @field numberOfCores number of cores for parallel computation
+#' @field quantileVec vector of quantiles to be calculated
+#' @field variableParameterPaths vector of paths of parameters to vary when performing sensitivity analysis
+#' @field pkParameterSelection list of selected PK parameters for sensitivity analysis
+PopulationSensitivityAnalysisSettings <- R6::R6Class(
```
msevestre: This feels kinds of weird to have a public member and then a function that updates the value

I'd suggest to change the implementation to use active keyword
Example (albeit complicated because we are wrapping .NET object here)
https://github.com/Open-Systems-Pharmacology/OSPSuite-R/blob/develop/R/pk-parameter.R#L49

It's also described in details in the R6 class chapter from Wickham

R/population-simulation-settings.R
```diff
@@ -0,0 +1,44 @@
+#' @title PopulationSimulationSettings
+#' @description  R6 class for Population Simulation Settings
+#' @field numberOfCores number of cores for parallel computation
+#' @field showProgress
+PopulationSimulationSettings <- R6::R6Class(
```
msevestre: Is there a reason why you are not using it the core object directly ? SimulationRunOptions

R/population-simulation-task.R
```diff
@@ -0,0 +1,81 @@
...
+        dir.create(file.path(self$workflowFolder, self$outputFolder))
+      }
+
+      for (set in structureSets) {
+        logWorkflow(
+          message = paste0("Run simulation: ", set$simulationSet$simulationName),
+          pathFolder = self$workflowFolder
+        )
+        if (self$validateInput()) {
+          taskResults <- self$getTaskResults(
```
msevestre: This indentation looks kind of weird. Is this how the styler formatted the code?

R/utilities-simulation.R
```diff
@@ -1,53 +1,146 @@
...
+  simRunOptions <- ospsuite::SimulationRunOptions$new(showProgress = FALSE)
+  simulationResult <- ospsuite::runSimulation(simulation, population = population, simulationRunOptions = simRunOptions)
+
+  logDebug(
+    message = paste0(ifnotnull(nodeName, paste0(nodeName, ": "), ""),"Simulation run complete"),
+    file = debugLogFileName,
+    printConsole = FALSE
+  )
+
+  ospsuite::exportResultsToCSV(
```
msevestre: SAme here. Line 84 is one line, line 92 is one parameter per line. I think we need to come up with a consistent way of writing the code. I don't think this is something that we have addressed in the styleguide.

R/population-simulation-settings.R
abdullahhamadeh: The core object takes numberOfCoresToUse whereas the reporting engine one takes numberOfCores, which will be potentially used by mpi

R/population-simulation-task.R
abdullahhamadeh: fixed

R/utilities-simulation.R
abdullahhamadeh: fixed

R/population-simulation-settings.R
msevestre: Hum... isn't it just a naming thing? The object are the same otherwise.

abdullahhamadeh: Yes I can change RE to interpret numberOfCoresToUse as the numberOfCores for mpi.  The idea was to make sure _core_ can't read this new object (if input by mistake) which could cause it to set numberOfCoresToUse >1 in a parallel run, where we want numberOfCoresToUse = 1 and numberOfCores (for mpi) to be >1.  Can remove it if you think this is overkill.

msevestre: I think each node when using mpi will have one core anyways so the .NET runner will use the max number of core in the node at any rate which would be one. 
Let's leave it for now but if the role of this object is nothing more than being an exact copy of the CORE object, we should think about using it

R/population-simulation-task.R
msevestre: Sorry if I was not clear. I meant the fact that  you write
```

          self$saveResults(
            set,
            taskResults
          )
``` 

instead of 
```
          self$saveResults(set,taskResults)
```

The former is not consistent with most code we have in the reporting engine

PR #109, opened 2020-03-30
102 funny logs

pchelle:


R/workflow.R
```diff
@@ -42,7 +42,7 @@ Workflow <- R6::R6Class(
       dir.create(self$workflowFolder, showWarnings = FALSE)
 
       logWorkflow(
-        message = self$reportingEngineInfo$print(),
+        message = private$reportingEngineInfo$print(),
```
msevestre: Usually, private member start with a .
so `private$.reportingEngineInfo`

tests/dev/script-mean-model-workflow-existing-results.R
```diff
@@ -0,0 +1,19 @@
+rm(list = ls())
+library(ospsuite.reportingengine)
```
msevestre: @pchelle  Simple :)

tests/dev/script-mean-model-workflow-existing-results.R
```diff
@@ -0,0 +1,19 @@
+rm(list = ls())
```
msevestre: @pchelle What does this line do? Just curious

R/workflow.R
pchelle: Ok, I'll make sure all these private fields start with "."

tests/dev/script-mean-model-workflow-existing-results.R
pchelle: This line clears the current working environment.
I use it to make sure that previous tests won't interfere with the current script.
Now that you point this out, it is definitely not necessary for the package and should be removed

PR #125, opened 2020-04-06
110 time profile

pchelle:
WIP:
Fixes most of the previous issues with time profile plots (using the last pull request from tlf)
- axes labels are fixed
- legend is a work in progress but better than before
- simulations are solid lines
- multiple simulations are allowed (I will add examples and tests ASAP)

Report is saved as markdown file, I will add documentation if the user wants to convert .md directly from R


msevestre:
@pchelle Build is failing
Also don't we need a PR in the TLF as well?

pchelle:
I did the PR for TLF last week, but we haven't merged it yet, do you think it is the reason why the build fails ?

For reporting engine, the PR is failing where I added tests today: 
Error: (unknown) (@test-tasks-in-workflow.R#7)
however, the tests where passing on my laptop
I am finish updating the vignette and I'll fix the issue, worse case, I'll remove the unit test

msevestre:
Sorry I missed the tlf PR Let me check

and probably. The tlf is copied in the tools folder to be installed by the script
so you need to  copy the same version 
https://github.com/Open-Systems-Pharmacology/OSPSuite.ReportingEngine/tree/develop/tools

It might not be the reason it fails but ...

msevestre:
ok I have just merged TLF
Once the package is updated, can you please download it from appveyor and copy it in the tools folder?

pchelle:
Awesome.
Thanks, I will.
The vignette was also failing at the same spot: initialization of a SimulationSet... with the following message: Error: object 'getCurrentConvertedObject' not found.
Might be because I am using a pkml that does not exist for the test (even if I don't use it).
Loading library(ospsuite) prevents it, but I have no idea why

msevestre:
>Might be because I am using a pkml that does not exist for the test (even if I don't use it).

What does this mean? Can you point me to the test?

msevestre:
something is weird here
https://github.com/Open-Systems-Pharmacology/OSPSuite.ReportingEngine/blob/develop/R/simulation-set.R#L83 (empty if?)

msevestre:
@pchelle This is weird. There is a test that instantiate the ospsuite. Wonder which one it is

msevestre:
Can you update the appveyor.yml file to use the right version? 
I'd do it but I have no access to your PR

msevestre:
@pchelle This line here
https://github.com/Open-Systems-Pharmacology/OSPSuite.ReportingEngine/blob/develop/appveyor.yml#L33

PR #135, opened 2020-04-19
38 sensitivity plot

pchelle:
WIP:
Parameter selection, cut-offs and display names needs to be integrated


R/calculate-pk-parameters-task.R
```diff
@@ -18,17 +18,47 @@ CalculatePKParametersTask <- R6::R6Class(
...
-    #' @param pkParameterResultsFilePath files where PK parameters are saved
-    #' @param generatedResultFileNames name of files where PK parameters are saved
-    #' @param ... parameters inherited from R6 class `Task` object
-    #' @return A new `CalculatePKParametersTask` object
-    initialize = function(...) {
-      super$initialize(...)
+    #' Save results from task run.
+    #' @param set R6 class `SimulationStructure`
+    #' @param taskResults list of results from task run.
+    saveResults = function(set,
```
msevestre, line 24: This identation is really messed up. 
Is there a reason why often you never put two parameter on the same line? 
The same question I also asked @abdullahhamadeh 
I understand that assignments are on multiple lines but this is just too much here

Can we just make sure that the function of a parameter call are on one line unless you go over the standard width of 120 col or whatever is the default and then you can align them? 

so for example 
```
 self$saveResults(
            set,
            taskResults
          )`
```
becomes

```
 self$saveResults(set,taskResults)
```

R/mean-model-workflow.R
```diff
@@ -286,72 +289,23 @@ MeanModelWorkflow <- R6::R6Class(
         message = "Starting run of mean model workflow",
         pathFolder = self$workflowFolder
       )
-      
-      if(self$resetReport$active){
+
+      if (self$resetReport$active) {
         resetReport(self$reportFileName,
```
msevestre, line 294: same here. Even in the same code it is not consistent

PR #136, opened 2020-04-20
update pop workflow

abdullahhamadeh:
changes made to update pop workflow with latest version of workflow class


Yuri05:
@abdullahhamadeh Tried the code from this PR with the example here: 
[Test04.zip](https://github.com/Open-Systems-Pharmacology/OSPSuite.ReportingEngine/files/4501893/Test04.zip)
(to reproduce: adjust the path in `Input/R_script.R` and start `R_script.R`)
Get the error: `Error in if (self$resetReport$active) { : argument is of length zero` (same as with RE 1.0.19).
Any ideas?

Yuri05:
@abdullahhamadeh Actually I get the same error as above when trying to execute `test-nonparallel-population.R` from this PR.

PR #137, opened 2020-04-20
Load paths in simulations before tasks

abdullahhamadeh:
Added functions to load pathID to simulation ahead of calculations


R/population-workflow.R
```diff
@@ -34,6 +34,7 @@ PopulationWorkflow <- R6::R6Class(
       super$initialize(...)
 
       # TO DO: include task parameters from initialization ?
+      self$resetReportSettings()
       self$simulatePopulationSettings()
```
msevestre, line 38: This has nothing to do with this PR but I am having trouble with the naming

For example simulatePopulationSettings does not `simulate` anything. Rather it create a task that will later be run correct? Should not we call these methods something different so that the non expert reader of the code does not believe that the simulation is going to be run right away?

/cc @pchelle @abdullahhamadeh

R/utils.R
```diff
@@ -162,3 +162,17 @@ getFileExtension <- function(filePath) {
...
 }
+
+
+#' @title loadSimulationWithUpdatedPaths
+#' @param simulationSet simulation set containing path to simulation file and pathIDs for quantities to be loaded into simulation object
+#' @return simulation object with pathIDs updated from simulationSet
+loadSimulationWithUpdatedPaths <- function(simulationSet) {
+  sim <- ospsuite::loadSimulation(filePath = simulationSet$simulationFile)
+  sim$outputSelections$clear()
+  quantities <- getAllQuantitiesMatching(paths = simulationSet$pathID, container = sim)
```
msevestre, line 173: use the `addOutputs` method from ospsuite-R
```
#' paths <- c("Organism|VenousBlood|Plasma|Caffeine", "Organism|ArterialBlood|**|Caffeine")
#' addOutputs(paths, sim)
```

R/workflow.R
```diff
@@ -19,11 +19,11 @@ Workflow <- R6::R6Class(
     #' @return A new `Workflow` object
     initialize = function(simulationSets,
                           workflowFolder) {
-      
+
       private$.reportingEngineInfo <- ReportingEngineInfo$new()
       # Check workflow folder input:
       # If workflow folder already exist throw a warning indicating the folder can be overwritten
-      self$workflowFolder <- workflowFolder %||% getwd()
+      self$workflowFolder <- workflowFolder %||% getwd()   #DO WE NEED THE %||% getwd() if workflowFolder is a required input?
```
msevestre, line 26: Agreed. Not required. Please remove
However VALIDATE that the workflowFolder is defined in the method using validateIsString() or sthg like this

R/utils.R
abdullahhamadeh: done

R/population-workflow.R
msevestre: I'd like feedback on this even though I am going to merge. 
@abdullahhamadeh

R/workflow.R
msevestre: @abdullahhamadeh This needs to be updated

abdullahhamadeh: done

R/population-workflow.R
abdullahhamadeh: we should create an issue for this because the mean model simulation task has a similar naming

msevestre: Sure. But do you concur that the naming is misleading OR is it me who does not understand the purpose of the functions?

abdullahhamadeh: I agree, I think they should be simulationTaskSettings  and populationSimulationTaskSettings

abdullahhamadeh: Also the task should be simulationTask and populationSimulationTask

msevestre: I'd even go as far as saying it should be something like `create` or `init`
for example createSimulationTask or sthg like this

PR #154, opened 2020-04-24
20 population pk parameters

pchelle:
WIP
Ratios and Pediatric options to be handled


Yuri05:
@pchelle 

>Between which population or parameters do we need to calculate the ratio ?

For every PK-parameter and for every population which is not the reference population.
s. e.g. https://github.com/Open-Systems-Pharmacology/Reporting-Engine/blob/develop/advanced_examples/05_Population_RatioComparison_Raltegavir/figures/pKParameter_Appendix.docx chapter 1.2). You just calculate the ratios `"Non-Reference-Population.Percentile_X / Reference-Population.Percentile_X"` for all percentiles (5, 25, 50, 75, 95) and plot it as a BW-plot
grafik

R/population-workflow.R
```diff
@@ -292,3 +312,21 @@ PopulationWorkflow <- R6::R6Class(
...
+
+#' @export
+PopulationWorkflowTypes <- enum(c(
+  "pediatric",
+  "parallelComparison",
+  "ratioComparison"
+))
+
+#' @export
+DemographyPaths <- enum(c(
```
msevestre, line 325: This is already available in ospsuite-r
https://github.com/Open-Systems-Pharmacology/OSPSuite-R/blob/c6dea52b9edb7509182fa67316cd9ae481d3b453/R/standard-path.R#L18

If something is missing, let's add. But let's not duplicate

R/utilities-calculate-pk-parameters.R
```diff
@@ -55,3 +55,354 @@ plotMeanPKParameters <- function(structureSet,
...
+#' @param workflowType workflowType Type of population workflow.
+#' Use enum `PopulationWorkflowTypes` to get list of workflow types.
+#' @param xParameters list of parameters to be plotted along x axis
+#' @param yParameters list of parameters to be plotted along y axis
+#' @return list of plots and tables with summary of PK parameters
+#' @export
+#' @import ospsuite
+#' @import tlf
+#' @import ggplot2
+plotPopulationPKParameters <- function(structureSets,
```
msevestre, line 74: This is a very, very complicated function!
* 250 lines
*  lots of loops: 3 or 4 imbricated for (hard to tell because I have to scroll so much)

Let split it in meaningful chunks

pchelle:
I don't understand why appveyor can't build ?
I just ran a devtools::check() on my computer and it worked

Yuri05:
>I don't understand why appveyor can't build ?

Because of the conflict in utils.R?

Yuri05:
@pchelle Can you add a commit to trigger a new AppVeyor-Build? Not sure what is wrong now.

msevestre:
Going to merge this as discussed this morning.

R/utils.R
```diff
@@ -174,9 +174,20 @@ loadSimulationWithUpdatedPaths <- function(simulationSet) {
...
+    }
   return(sim)
 }
+
+#' @title lastPathElement
+#' @param path simulation path
+#' @return last path element as character string
+#' @export
+lastPathElement <- function(path){
+  return(sub(pattern = "^.*[|]",
```
msevestre, line 190: @Yuri05 Hum. I'd suggest to use the `toPathArray` method in get the last element 

https://github.com/Open-Systems-Pharmacology/OSPSuite-R/blob/develop/R/utilities-path.R#L12

tests/dev/03_Population_Pediatric_Raltegavir/test-example-03.R
```diff
@@ -0,0 +1,31 @@
+#------------------------------
+# Advanced example 03: Set/run population workflow
+#------------------------------
+
+library(ospsuite.reportingengine)
+
```
msevestre, line 6: This looks very nice from the API point of view.

PR #156, opened 2020-04-26
Add pop sensitivity plots

abdullahhamadeh:


pchelle:
Appveyor seems to fail to install ospsuite-R latest version

pchelle:
https://github.com/Open-Systems-Pharmacology/OSPSuite.ReportingEngine/blob/3df7e5093cf72302587431801a12c639c3105c8b/R/utilities-sensitivity-analysis.R#L640

@abdullahhamadeh
I think you need to document import ggplot2 or/and call ggplot2 functions like this ggplot2::aes_string, ggplot2::geom_vline,   otherwise devtools::check will not pass

pchelle:
Also can you replace ggplot2::ggplot() by tlf::initializePlot(plotConfiguration)
with plotConfiguration an input of the function and from plotPopulationSensitivity, you can call the method with plotConfiguration = settings$plotConfiguration.
This way users will have the ability to modify the plotConfiguration settings for this plot

msevestre:
> Appveyor seems to fail to install ospsuite-R latest version

This is an issue with the R version being installed. A new major was just released 4.0 which breaks the build. We will have to enforce 3.6.... just need to find how :)

msevestre:
ok so I have managed to fix the build issues but I think there is another issue here with a doc file not compiling. I'll try to fix this but if I do not succeed, @abdullahhamadeh you'll have to check it

msevestre:
@abdullahhamadeh @pchelle 
This is the error here
It can be reproduce by running the check locally
```
 creating vignettes (34.7s)
   Quitting from lines 155-156 (mean-model-workflow.Rmd) 
   Error: processing vignette 'mean-model-workflow.Rmd' failed with diagnostics:
   unused arguments (metaData = metaData, plotConfiguration = plotConfiguration)
   Execution halted
Error in (function (command = NULL, args = character(), error_on_status = TRUE,  : 
  System command error
Calls: <Anonymous> ... with_envvar -> force -> do.call -> <Anonymous> -> throw
Execution halted
```

Can you please fix this and then this should be good.

msevestre:
There is also a warning regarding addFigureChunk documentation. Not sure what it is all about

Also all those utilities method are exported. Is this on purpose?

msevestre:
Seems that there are still some issues with the build. ... F$#^%$#^ing R

msevestre:
success! there are still a bunch of warnings that should be dealt with
I'll merge

PR #166, opened 2020-04-29
Parallel error handling

abdullahhamadeh:


R/utilities-mpi-error-handling.R
```diff
@@ -0,0 +1,153 @@
+#' @title checkSimulationLoaded
+#' @description Check that the simulation has been loaded on the core.
+#' @param simulation the simulation object to be checked if loaded successfully
+#' @param simulationFilePath the path to the simulation object to be checked if loaded successfully
+#' @export
+checkSimulationLoaded <- function(simulation, simulationFilePath) {
```
msevestre, line 6: I don't understand this method. What does that mean loaded?
How can you have an instance of a simulation that is not loaded? It is not possible to construct a simulation otherwise

msevestre: From a code perspective, this can be simplified by returning 

```R
checkSimulationLoaded <- function(simulation, simulationFilePath) {
   identical(simulation$sourceFile, simulationFilePath)
}
```

or 
```R
checkSimulationLoaded <- function(simulation, simulationFilePath) {
  return( identical(simulation$sourceFile, simulationFilePath))
}
```

R/utilities-mpi-error-handling.R
```diff
@@ -0,0 +1,153 @@
...
+    success <- TRUE
+  }
+  return(success)
+}
+
+#' @title checkLibraryLoaded
+#' @description Check that a library has been successfully loaded on the core.
+#' @param libraryName the library to be checked if loaded successfully
+#' @export
+checkLibraryLoaded <- function(libraryName) {
```
msevestre, line 18: Same here. This can be greatly simplified returning the boolean from the check

R/utilities-mpi-error-handling.R
```diff
@@ -0,0 +1,153 @@
...
+}
+
+
+#' @title checkNotNull
+#' @description Check that an object is not null.
+#' @param object to be checked if not null
+#' @export
+checkNotNull <- function(object) {
+  success <- FALSE
+  if (!is.null(object)) {
```
msevestre, line 33: Why do we need a method to replace a core method?

R/utilities-mpi-error-handling.R
```diff
@@ -0,0 +1,153 @@
...
+    success <- TRUE
+  }
+  return(success)
+}
+
+#' @title checkAllCoresSuccessful
+#' @description Check that all cores executed an mpi.remote.exec command successfully.
+#' @param coreResults list of results returned by each core after an mpi.remote.exec command is complete
+checkAllCoresSuccessful <- function(coreResults) {
+  allCoreResults <- sapply(1:length(coreResults), function(n, obj) obj[[n]], coreResults)
```
msevestre, line 43: Can you give an example of how this will be called? I don't understand the syntax

R/utilities-mpi-error-handling.R
```diff
@@ -0,0 +1,153 @@
...
+    logErrorThenStop(message = "Individual parameters updated successfully on all cores.", logFolderPath = logFolder)
+  }
+}
+
+
+#' @title verifySimulationRunSuccessful
+#' @description Check that all cores ran simulation successfully
+#' @param simulationRunSuccess logical vector indicating success of simulation run on all cores
+verifySimulationRunSuccessful <- function(simulationRunSuccess, logFolder = getwd()) {
+  success <- FALSE
```
msevestre, line 102: No need to initialize to FALSE since it's reset right after

R/utilities-mpi-error-handling.R
```diff
@@ -0,0 +1,153 @@
...
+  } else {
+    logErrorThenStop(message = "Sensitivity analyses not completed successfully on all cores.", logFolderPath = logFolder)
+  }
+}
+
+#' @title verifyAnyPreviousFilesRemoved
+#' @description Check that any existing partial sensitivity results from individual cores have been removed
+#' @param sensitivityRunSuccess logical vector indicating success of result file removal
+verifyAnyPreviousFilesRemoved <- function(anyPreviousPartialResultsRemoved, logFolder = getwd()) {
+  success <- FALSE
```
msevestre, line 131: No need to initialize to FALSE since it's reset right after

R/utilities-sensitivity-analysis.R
```diff
@@ -44,15 +44,10 @@ runSensitivity <- function(structureSet,
   # Otherwise sensitivity analysis will be run on master core only.
   if (numberOfCores > 1) {
     Rmpi::mpi.spawn.Rslaves(nslaves = numberOfCores)
-    Rmpi::mpi.remote.exec(library("ospsuite"))
-    Rmpi::mpi.remote.exec(library("ospsuite.reportingengine"))
-
-    # Pass to cores the inputs that will not change from one individual to another
+    Rmpi::mpi.remote.exec("HERE")
```
msevestre, line 47: REMOVE

R/utilities-simulation.R
```diff
@@ -226,9 +254,10 @@ runParallelPopulationSimulation <- function(structureSet,
 #' @import ospsuite
 updateSimulationIndividualParameters <- function(simulation, individualParameters = NULL) {
   if (is.null(individualParameters)) {
-    return()
+    return(TRUE)
   }
   ospsuite::setParameterValuesByPath(parameterPaths = individualParameters$paths, values = individualParameters$values, simulation = simulation)
```
msevestre, line 259: Why the return values here? This only makes sense if you try catch the call to ospsuite right? Otherwise it's always true or throws an exception

R/utilities-mpi-error-handling.R
```diff
@@ -0,0 +1,153 @@
...
+#' @param coreResults list of results returned by each core after an mpi.remote.exec command is complete
+checkAllCoresSuccessful <- function(coreResults) {
+  allCoreResults <- sapply(1:length(coreResults), function(n, obj) obj[[n]], coreResults)
+  return(all(allCoreResults))
+}
+
+#' @title loadSimulationOnCores
+#' @description Send structureSet to core, check its simulation has been loaded successfully
+#' @param structureSet containing simulationSet which contains path to simulation file and pathIDs to be loaded in simulation object as outputs
+loadSimulationOnCores <- function(structureSet, logFolder = getwd()) {
```
msevestre, line 50: I don't think we should have default value for logFolder in utilities method. This is a mandatory parameter

R/utilities-mpi-error-handling.R
```diff
@@ -0,0 +1,153 @@
...
+    logWorkflow(message = paste("Simulation file", structureSet$simulationSet$simulationFile, "loaded successfully on all cores"), pathFolder = logFolder)
+  } else {
+    logErrorThenStop(message = paste("Simulation file", structureSet$simulationSet$simulationFile, "not loaded successfully on all cores"), logFolderPath = logFolder)
+  }
+}
+
+#' @title loadSimulationOnCores
+#' @description Send libraryName to core, load the library and check that it has been loaded successfully
+#' @param libraryName string containing name of library to be loaded
+loadLibraryOnCores <- function(libraryName, logFolder = getwd()) {
```
msevestre, line 67: I don't think we should have default value for logFolder in utilities method. This is a mandatory parameter

R/utilities-mpi-error-handling.R
```diff
@@ -0,0 +1,153 @@
...
+  success <- FALSE
+  Rmpi::mpi.bcast.Robj2slave(obj = libraryName)
+  Rmpi::mpi.remote.exec(library(libraryName, character.only = TRUE))
```
msevestre, line 70: What does  character.only = TRUE means?

R/utilities-mpi-error-handling.R
```diff
@@ -0,0 +1,153 @@
...
+  } else {
+    logErrorThenStop(message = paste(libraryName, "not loaded successfully on all cores"), logFolderPath = logFolder)
+  }
+}
+
+#' @title loadSimulationOnCores
+#' @description Send libraryName to core, load the library and check that it has been loaded successfully
+#' @param libraryName string containing name of library to be loaded
+updateIndividualParametersOnCores <- function(individualParameters, logFolder = getwd()) {
+  success <- FALSE
```
msevestre, line 85: No need to initialize to FALSE since it's reset right after

R/utilities-mpi-error-handling.R
abdullahhamadeh: mpi.remote.exec sends a command to each core and returns the combined results of the command from each core in a list.  I modified the methods that are called by mpi.remote.exec on each core to return true if successfully completed.   This function just checks that all returned values in the list are TRUE.

msevestre: I get what it does. I don't understand the inplemenation. Sapply works with an iterrable directly. Why do you need to do 1: lengths to retrieve each single item?

R/utilities-mpi-error-handling.R
abdullahhamadeh: I want to check that `sim` is loaded correctly on each core.  If I initialize `sim `to NULL on the core, then load the simulation to `sim `on the core, then I can either
1) check that `sim` is not null, or
2) perform the above check (ie that the simulation was loaded using the correct sourceFile

is 2 overkill?

R/utilities-mpi-error-handling.R
abdullahhamadeh: library(libraryName) would attempt to load a library called "libraryName".   character.only = TRUE tells it to interpret libraryName as a variable that stores a string.

R/utilities-simulation.R
abdullahhamadeh: updateSimulationIndividualParameters is used in updateIndividualParametersOnCores to verify and log that the parameters were loaded on each core.

R/utilities-mpi-error-handling.R
abdullahhamadeh: Done

R/utilities-mpi-error-handling.R
abdullahhamadeh: Removed and replaced with !is.null()

R/utilities-mpi-error-handling.R
abdullahhamadeh: Simplified this

R/utilities-mpi-error-handling.R
abdullahhamadeh: Removed

R/utilities-mpi-error-handling.R
abdullahhamadeh: Removed

R/utilities-mpi-error-handling.R
abdullahhamadeh: Removed

R/utilities-mpi-error-handling.R
abdullahhamadeh: Removed

R/utilities-mpi-error-handling.R
msevestre: oh i see.  
Wouldn't it crash if simulation is null?

R/utilities-mpi-error-handling.R
```diff
@@ -0,0 +1,130 @@
...
+
+
+#' @title checkAllCoresSuccessful
+#' @description Check that all cores executed an mpi.remote.exec command successfully.
+#' @param coreResults list of results returned by each core after an mpi.remote.exec command is complete
+checkAllCoresSuccessful <- function(coreResults) {
+  allCoreResults <- sapply(coreResults, function(obj) {
+    obj
+  })
+  return(all(allCoreResults))
```
msevestre, line 26: Just curious. Wouldn't all(allCoreResults) work?

msevestre: sorry all(coreResults)?

R/utilities-mpi-error-handling.R
msevestre: ok cool

PR #173, opened 2020-05-01
Fixes 126 MatlabID -> ID

pchelle:

PR #178, opened 2020-05-04
174 simulation set

pchelle:


Yuri05:
@pchelle Please let me know when we can merge this. Another question: is PR #173 not required then? (because I see "Fixes 126 MatlabID -> ID" in the commit history of this PR)

pchelle:
Yes, it can be merged.
Since it includes the commits from PR#173, PR#173 should not be required anymore.
I tried the code on Test06, it seems to work but it needs to be updated with the new output class.
Let me know if I should send the updated code as a reply of the corresponding issue.

Yuri05:
@pchelle Cool. I've closed #173 
>Let me know if I should send the updated code as a reply of the corresponding issue.

Yep, that would be great.

PR #187, opened 2020-05-16
Fixes #24 Framework for population study design

### LINKED ISSUE #24: Population study design
Sometimes it's required to add a special parameters variation which is not part of the population table (e.g. not supported by PK-Sim). E.g.:
* Set dose per body weight **range** (e.g. "`1 mg for BW in [10kg..20kg); 2 mg for BW>20kg`")
* Set dose per age **range**
* Set dose per BSA **range**

For that, additional "study design" table is provided (current template for the Matlab implementation s. [here](https://github.com/Open-Systems-Pharmacology/Reporting-Engine/blob/develop/src/code/templates/StudyDesign.xlsx))

Population table is then extended according to the design study ([current Matlab implementation](https://github.com/Open-Systems-Pharmacology/Reporting-Engine/blob/develop/src/code/engine_library/readStudyDesign.m))

pchelle:
WIP: 
I need to connect this new frame work into workflows and to add a few tests in test-that
In the current state of this frame work, the user will need to provide PK-Sim base units (since the simulation object is required to check that, I may include simulation only in a later version)


pchelle:
The PR is good for review now

tests/dev/03_Population_Pediatric_Raltegavir/StudyDesign.csv
```diff
@@ -0,0 +1,8 @@
+Organism|Weight,Organism|Weight,Gender,App1|Drugmass,App2|Drugmass
+kg,kg,,mg,mg
+SOURCE_MIN,SOURCE_MAX,SOURCE_EQUALS,TARGET,TARGET
+20,40,1,2,3
+20,40,2,2.5,3.5
+40,60,1,10,12
+40,60,2,14,16
+40,,,20,20
```
Yuri05, line 8: Interesting example. Actually it is invalid. Was the last line meant to be e.g. _**60**,,,20,20_ ?

R/population-study-design.R
```diff
@@ -0,0 +1,226 @@
...
+  }
+
+  return(values)
+}
+
+#' @title loadStudyDesign
+#' @description Load a StudyDesign object from a file containing a study design table
+#' In this table, Line 1 is path,  Line 2 is unit, Line 3 is type and subsequent lines are values.
+#' The cells for type can include "SOURCE" OR "TARGET". "SOURCE" type can also include subtype "MIN", "MAX" or "EQUALS".
+#' In the current version, unit is expected to be intern units of simulation.
```
Yuri05, line 51: Do I understand t right: at the moment the units in a design study files are ignored and the values in a design study file must be always in the internal OSP units (e.g. age always in years etc.)?
If so: why?

tests/dev/03_Population_Pediatric_Raltegavir/StudyDesign.csv
pchelle: It should indeed be 60.

R/population-study-design.R
pchelle: Yes. It is currently ignored.
At the moment, only a simulation object has the knowledge and concept of unit and dimension.

So there can be 2 ways of dealing with unit: 
1) Add the concept in a population object (this would make a lot of changes in the ospsuite package)
2) Include the simulation object to the function addStudyParameter (preferred way)
Use the simulation object to check if added Study Parameter exists in the simulation and that its unit/dimension exist.
If you have an actual example to try that out, I may be able to include that in the current PR.

Yuri05: Would it be enough to convert every column to the base unit of its corresponding dimension?
(@msevestre Does ospsuite-R provide such a possibility? )

Then the simulation object would not be required.

pchelle: I see, the function `toBaseUnit` is what you are referring to.
The process would then be, getDimensionFor study design unit, and convert unit toBaseUnit.
I can add that process to fix the unit.
In the end, I still think we'll need the simulation object at some point since the extended population is meant to be passed on the simulation run (so we can check before the run that paths and their units match and are correctly defined).

pchelle: I just tried to implement and it does not work: `toBaseUnit` requires a quantity object which is derived from a simulation object.
I think including the simulation directly is the best solution.
@Yuri05, is it okay if I through an error if TARGET paths do not exist ?

pchelle: *throw

Yuri05:
>is it okay if I throw an error if TARGET paths do not exist ?

Yep, it's an error

tests/dev/03_Population_Pediatric_Raltegavir/StudyDesign.csv
msevestre: Why is it invalid ?

It does not make sense but it is still valid. In that case the the last entry would win right? 20 would be the value for all drug mass

R/population-study-design.R
msevestre: I can update the code in toBaseUnit to be similar to what `toUnit` does and support a dimension name as well. In that case, the simulation would not be required

But you would not check that the path exists. 
I have a few updates to do to ospsuite-r so I'll do that today and then you can update the PR

R/population-study-design.R
```diff
@@ -0,0 +1,226 @@
+#' @title addStudyParameters
+#' @param population `Population` object
+#' @param studyDesignFile file name of study design table
+#' @export
+#' @import ospsuite
+addStudyParameters <- function(population, studyDesignFile = NULL) {
```
msevestre, line 6: I am a bit confused with the signature. If someone calls addStudyDesign, shouldn't the file exist?
Or do you intend to call it no matter what?

R/population-study-design.R
```diff
@@ -0,0 +1,226 @@
...
+    population$setParameterValues(parameterPath, updatedTargetValues)
+  }
+}
+
+#' @title updateTargetValues
+#' @param values initial vector to update
+#' @param targetValues vector of values to be assigned using `sourceExpressions`
+#' @param sourceExpressions study design expressions to be evaluated
+#' @param data population data as data.frame
+#' @export
```
msevestre, line 31: Is this really a public method to export ? This looks like internal stuff to me

R/population-study-design.R
```diff
@@ -0,0 +1,226 @@
...
+#' In this table, Line 1 is path,  Line 2 is unit, Line 3 is type and subsequent lines are values.
+#' The cells for type can include "SOURCE" OR "TARGET". "SOURCE" type can also include subtype "MIN", "MAX" or "EQUALS".
+#' In the current version, unit is expected to be intern units of simulation.
+#' @param studyDesignFile file name of study design table
+#' @return A `StudyDesign` object`
+#' @export
+loadStudyDesign <- function(studyDesignFile) {
+  validateIsString(studyDesignFile)
+  designData <- read.csv(studyDesignFile, header = FALSE, stringsAsFactors = FALSE)
+
```
msevestre, line 58: you can safely remove all those spaces here between lines. There is no if etc.. so it should be all without extra space

R/population-study-design.R
```diff
@@ -0,0 +1,226 @@
...
+  public = list(
+    source = NULL,
+    targets = NULL,
+
+    #' @description Create a new `StudyDesign` object.
+    #' @param data data.frame read from study design file
+    #' @return `StudyDesign` class object
+    initialize = function(data) {
+      self$targets <- mapStudyDesignTargets(data)
+
```
msevestre, line 86: same here. No need for extra spaces between assignments

R/population-study-design.R
```diff
@@ -0,0 +1,226 @@
...
+      self$source <- mapStudyDesignSources(data)
+
```
msevestre, line 88: This one is ok to separate assignments from action

R/population-study-design.R
```diff
@@ -0,0 +1,226 @@
...
+  validateIsPositive(sum(sourceFilter))
+  # Enforce data.frame with drop = FALSE
+  sourceData <- data[, sourceFilter, drop = FALSE]
+
+  sourceExpressions <- NULL
+  for (columnIndex in seq(1, ncol(sourceData))) {
+    path <- sourceData[studyDesignPathLine, columnIndex]
+    unit <- sourceData[studyDesignUnitLine, columnIndex]
+    sourceType <- sourceData[studyDesignTypeLine, columnIndex]
+
```
msevestre, line 167: same here. Remove all those spaces between variable assignments

R/utilities-simulation.R
```diff
@@ -8,6 +8,7 @@ simulateModelForPopulation <- function(structureSet,
                                        settings = NULL,
                                        logFolder = getwd()) {
   population <- ospsuite::loadPopulation(structureSet$simulationSet$populationFile)
+  addStudyParameters(population, structureSet$simulationSet$studyDesignFile)
```
msevestre, line 11: ok I have answered my own question from before here

R/population-study-design.R
pchelle: From what I understand this function may be called mostly from within a workflow.
So, I was thinking of using the function no matter what the user provides to a simulation set.
But it may be better to split between an actual addStudyParameter and a workflow specific function that will used a simulation set as an input and will check if the file is provided.

R/population-study-design.R
pchelle: You are right, there are a couple others I should not export.
I tend to do this at the beginning to test the functions works correctly when updating the package. I'll remove them asap.

Yuri05:
@pchelle I'm a bit lost with the 2 PRs. 
Can we close "Tlf install fix" in favor of this one or do we need both PRs?

pchelle:
Yes. I am going to close "tlf install fix" which was just a way of testing what was wrong with the installation of tlf.
Sorry, I did not close it sooner, I had a few meetings today and haven't closed it yet.

Yuri05:
no problem, thx for clarification.
Any remaining TODOs in this PR or is good to merge now?

pchelle:
I think it is good to merge now.
@abdullahhamadeh , can you rebase your PRs on this one ?

abdullahhamadeh:
updated this pr with my changes.  closing #195 and #196.

PR #197, opened 2020-05-22
Tlf install fix

pchelle:


pchelle:
This construction seems to work, not exactly sure why.
R works in mysterious ways...

It seems that remote::install_github works while devtools::install_github does not work, 
but I don't know why and what would be the difference between them.

Issue #171, opened 2020-04-30
totalSensitivityThreshold: Should it be used when the list of variable parameters is predefined by the user?


msevestre:
It might make sense to only consider the parameter when the list of variable parameters is retrieved automatically from the simulation

If the list is preset by the user, the threshold parameter would be set to 1?

FYI @Yuri05

Yuri05:
@sfrechen @TWendl @Incei @lkuepfer @ThomasGaub please comment

Incei:
I dont understand the question...what do you mean with theshold?

Yuri05:
Threshold = "cut-off" used for plots of the most sensitive parameters (https://docs.open-systems-pharmacology.org/shared-tools-and-example-workflows/sensitivity-analysis#tab-sensitivity-analysis)

Incei:
I don't have a strong preference. Perhaps @KatrinCoboeken ?

KatrinCoboeken:
The threshold makes sense if the parameter list is very long. 
So my favorite option would be that user can change this manually, with default:
 use it for automatically parameter selection, 
 do not use it for user defined selection

Yuri05:
That means, that:
* the **default** for the cut-off would be:
  * 1 in case sensitivity parameters list was provided by user
  * 0.9 otherwise
* default cut-off can be overwritten by user in both cases

KatrinCoboeken:
yes

TWendl:
agree with @Yuri05 on threshold default settings!

abdullahhamadeh:
Is there a preferred default threshold for the population sensitivity plot, o

abdullahhamadeh:
Is there a different default for the population sensitivity plot, or 1 as well?

Yuri05:
>or 0.9/1 as well?

Yes.

abdullahhamadeh:
In the sensitivity results CSV file, do we want all sensitivity results or just those within the threshold?

Yuri05:
I would say: all. Plot just does not look good with too many points, but I don't see any reason why we should exclude this information from the csv.

abdullahhamadeh:
For now I am manually selecting the top X% most sensitive parameters, where X is the threshold.  Is there a way to use ospsuite totalSensitivityThreshold feature?  I would need a way to import to/export from a CSV SA result file the SA results falling within totalSensitivityThreshold.

msevestre:
What do you mean manually?

[...]

Issue #167, opened 2020-04-29
Sensitivity analyses


lkuepfer:
The layout of the output of sensitvity analyses (mean model) looks strange:

![Raltegravir 10 mg   (lactose formulation)-AUC_inf-1](https://user-images.githubusercontent.com/29250436/80648438-78731680-8a70-11ea-89a9-3864166ac9d4.png)

msevestre:
What's wrong with this ? (just kidding)

Can you attach the workflow and folder so that we can reproduce?

lkuepfer:
Freundliche Grüße / Best regards,

Dr. Lars Kuepfer
Scientist Systems Pharmacology

////////////////////

Bayer AG
Research & Development, Pharmaceuticals
Systems Pharmacology & Medicine II
Building B106, 106
51368 Leverkusen, Germany
Tel:       +49 214 30-22745
Mobile: +49 173 24-79088
E-mail:  lars.kuepfer@bayer.com
Web:     http://www.bayer.com
Datenschutzhinweise<https://www.bayer.de/de/datenschutzhinweise-fuer-ausgewaehlte-verarbeitungstaetigkeiten.aspx> // Data privacy information<https://www.bayer.com/en/data-privacy-information-for-specific-processing-activities.aspx>

/// Vorstand: Werner Baumann, Vorsitzender | Liam Condon, Wolfgang Nickl, Stefan Oelrich, Heiko Schipper
/// Vorsitzender des Aufsichtsrats: Werner Wenning
/// Sitz der Gesellschaft: Leverkusen | Amtsgericht Köln, HRB 48248

Von: Michael Sevestre <notifications@github.com>
Gesendet: Mittwoch, 29. April 2020 23:31
An: Open-Systems-Pharmacology/OSPSuite.ReportingEngine <OSPSuite.ReportingEngine@noreply.github.com>
Cc: Lars Kuepfer <lars.kuepfer@bayer.com>; Author <author@noreply.github.com>
Betreff: Re: [Open-Systems-Pharmacology/OSPSuite.ReportingEngine] Sensitivity analyses (#167)


What's wrong with this ? (just kidding)

Can you attach the workflow and folder so that we can reproduce?

Yuri05:
Closing in favor of #201 and #202

Issue #165, opened 2020-04-29
Population sensitivity plots: legends


Yuri05:
* Remove caption above the plot and "Parameter" on the left
* reduce font size for parameter names
* Legend should look like this (s. [Spec](https://github.com/Open-Systems-Pharmacology/OSPSuite.ReportingEngine/issues/20))
  * <PopulationSimulationSet_name_1> x_th percentile individual
  * <PopulationSimulationSet_name_1> y_th percentile individual
  * <PopulationSimulationSet_name_2> x_th percentile individual
  * <PopulationSimulationSet_name_2> y_th percentile individual
![grafik](https://user-images.githubusercontent.com/25061876/80583454-809b6980-8a10-11ea-903b-8361b1114ba7.png)

Yuri05:
Fixed in the meanwhile (new issue for the font size created)

Issue #164, opened 2020-04-29
Pop sensitivity plots with several populations


Yuri05:
[Sensi.zip](https://github.com/Open-Systems-Pharmacology/OSPSuite.ReportingEngine/files/4550934/Sensi.zip)
2 populations are used, however sensitivity plots show only 1 value for each percentile
![Raltegravir 10 mg   (lactose formulation)-Raltegravir Adult Population-C_max-Organism-Heart-Plasma-Raltegravir-Concentration](https://user-images.githubusercontent.com/25061876/80582831-b0963d00-8a0f-11ea-8b6c-d9a507289117.png)

msevestre:
How should the plot look like? Isn't it going to be impossible to read?

Yuri05:
>Isn't it going to be impossible to read? 

For 2-3 pops should be ok, for 4 or more not really readable indeed. But: it's users problem then:)
In the current Matlab-RE it looks e.g. like this: https://github.com/Open-Systems-Pharmacology/Reporting-Engine/blob/develop/advanced_examples/07_Population_Pediatric_Raltegavir_inclSens/figures/sensitivity_Appendix.docx

Yuri05:
Fixed with #212

PR #198, opened 2020-05-22
Re add pop sensitivity and path check

abdullahhamadeh:
I can't see my changes from #195 and #196.   I have merged them with develop again.


Yuri05:
@abdullahhamadeh Tried this PR with [SensiCrash.zip](https://github.com/Open-Systems-Pharmacology/OSPSuite.ReportingEngine/files/4671544/SensiCrash.zip) and got the error message

```
[1] "23/05/2020 - 12:23:03 : Run sensitivity for simulation: Rifampin po, 450 mg, Bioequivalence - PopSim"
Error in data.frame(Outputs = outputColumn, pkParameters = pkParameterColumn,  :
  arguments imply differing number of rows: 85, 80
```

R/messages.R
```diff
@@ -84,8 +84,12 @@ messages <- list(
...
     paste0("Workflow type '", workflowType, "' requires one reference population, but no reference population was provided.")
   },
-  
-  errorNotADimension = function(values){
+
+  errorNotADimension = function(values) {
     paste0("Expected a dimension. Check that '", paste0(values, collapse = "', '."), "' is included in ospsuite::allAvailableDimensions().")
+  },
+
+  invalidOuputPath = function(pth, simName) {
```
Yuri05, line 92: typo: **pth**

R/utilities-sensitivity-analysis.R
```diff
@@ -569,45 +571,68 @@ plotTornado <- function(data,
...
+                                      logFolder = NULL,
+                                      settings,
+                                      workflowType = NULL,
+                                      xParameters = NULL,
+                                      yParameters = NULL) {
+  dFList <- list()
+  i <- 0
+  for (structureSet in structureSets) {
+    indexDf <- read.csv(file = structureSet$popSensitivityAnalysisResultsIndexFile)
+    rankFilter <- 12
```
Yuri05, line 591: Still a TODO here. s. https://github.com/Open-Systems-Pharmacology/OSPSuite.ReportingEngine/issues/171#issuecomment-624548277

abdullahhamadeh:
Fixes #171 
Fixes #167 
Fixes #165 
Fixes #164

Yuri05:
@abdullahhamadeh With this PR I get the error message in plot sensitivity task for pop.
```
[1] "24/05/2020 - 14:58:42 : Starting: Plot Sensitivity task in Alpha Testing"
Fehler in rClr::clrCall(self$ref, "AllPKParameterSensitivitiesFor", pkParameterName,  : 
  Type:    System.MissingMethodException
Message: Could not find a suitable instance method AllPKParameterSensitivitiesFor on type OSPSuite.Core.Domain.SensitivityAnalyses.SensitivityAnalysisRunResult for method parameters String, String, SymbolicExpression
```
I use latest ospsuite-R package version **9.0.76**
[SensiCrash_2.zip](https://github.com/Open-Systems-Pharmacology/OSPSuite.ReportingEngine/files/4673557/SensiCrash_2.zip)

abdullahhamadeh:
I will look at the mean sensitivity plot issue but for now this pr fixes the population sensitivity plot crash.  The cause of the crash was that the threshold wasn't being set to anything by default.  It had to be added explicitly.

I have fixed this so that we have the following way of setting the threshold upon initialization of the workfow: When the pop workflow is initialized it looks for whether a threshold was explicitly set for the pop sensitivity plot using the SensitivityPlotSettings class.  If not, it looks at the population sensitivity analysis settings.  If a set of perturbation parameter paths has been set there, the threshold is 1, else 0.9.  If the population sensitivity analysis settings are null, an instance of SensitivityPlotSettings class is used to set the theshold to 0.9 by default.

There is an issue currently, which is that if the sensitivity analysis is updated with a set of perturbation parameter paths _after_ the initialization of the workflow, the pop sensitivity plot task has no way of knowing that the threshold should be updated to 1 (instead of 0.9).  

Perhaps the simulation structure should also hold a copy of the task settings?  It's a situation where a pointer to the pertubation parameter paths would be extremely useful becuase then the threshold would automatically update depending on whether these paths are set.

Yuri05:
oki, merging for now, let's create separate issues for the remaining topics

PR #205, opened 2020-05-26
180 workflow report sections

pchelle:
WIP


Yuri05:
@pchelle It's getting better :) 
Question: why do some captions switch to a new line long before the page width is reached?
grafik

pchelle:
The reason is `pandoc`, the function that adds the table of content.
This function uses the 80 characters line break rule... consequently, whenever it was not inline markdown code for figures or tables, it was just breaking the line before reaching 80 characters.
I am pushing the option wrap=none to pandoc so it will leave the lines as is.

By the way, this function has a lot of options:
- I used a Github Flavoured Format (gfm) instead of plain markdown because it is closer to the writing we are used to. If you wondered why the option `format="gfm"`
- It is also supposed to number the sections, however it does not work for markdown formats... that's why I had to use an unefficient counting of sections and figures.

abdullahhamadeh:
@pchelle I am getting the error `pandoc.exe: Unknown writer: gfm`.  Is there a separate package for it?

pchelle:
The package is `knitr`.
In `knitr::pandoc`, "gfm" has replaced "github_document" from older versions.
Updating the package should work (if the package version is not in the okay list, I'll use github_document instead)

PR #212, opened 2020-06-04
Add missing pop sa plot points and use output paths for sa

abdullahhamadeh:


R/utilities-sensitivity-analysis.R
```diff
@@ -403,30 +386,78 @@ runPopulationSensitivityAnalysis <- function(structureSet, settings, logFolder =
...
+#' @description Read PK parameter results into a dataframe and set QuantityPath,Parameter and Unit columns as factors
+#' @param structureSet `SimulationStructure` R6 class object
+#' @return pkResultsDataFrame, a dataframe storing the contents of the CSV file with path pkParameterResultsFilePath
+#' @import ospsuite
+getPKResultsDataFrame <- function(structureSet) {
+  pkResultsDataFrame <- read.csv(structureSet$pkAnalysisResultsFileNames,
+    encoding = "UTF-8",
+    check.names = FALSE,
+    stringsAsFactors = FALSE
+  )
```
pchelle, line 399: Why not using opsuite::importPKAnalysesFromCSV + opsuite::pkAnalysesAsDataFrame intead ?

R/utilities-sensitivity-analysis.R
```diff
@@ -403,30 +386,78 @@ runPopulationSensitivityAnalysis <- function(structureSet, settings, logFolder =
...
+        function(y) {
+          y$pkParameter
+        }
+      ))
+    } else {
+      # case where no pkParameters are specified for an output and SA performed for all pk parameters by default
+      pkResultsDataFrameThisOutput <- pkResultsDataFrame[ pkResultsDataFrame$QuantityPath == op$path, ]
+      pkParameterNamesEachOutput[[op$path]] <- unique(pkResultsDataFrameThisOutput$Parameter)
+    }
+  }
```
pchelle, line 419: You can simplify the loop as below:
`for (output in structureSet$simulationSet$outputs){ if(!is.null(output$pkParameters)) ... }`

Besides, for character type variables, I think `pkResultsDataFrame$QuantityPath %in% op$path` is better than `pkResultsDataFrame$QuantityPath == op$path`

Would it also be better to start the loop with the `else` content and then only use an `if` to filter and check that the user selected pkParameters are in pkResultsDataFrame ?

R/utilities-sensitivity-analysis.R
```diff
@@ -403,30 +386,78 @@ runPopulationSensitivityAnalysis <- function(structureSet, settings, logFolder =
...
   pkParameterColumn <- NULL
   individualIdColumn <- NULL
   valuesColumn <- NULL
   unitsColumn <- NULL
   quantileColumn <- NULL
   for (output in outputs) {
+    pkParameters <- unique(allPKResultsDataframe$Parameter[allPKResultsDataframe$QuantityPath == output])
     for (pkParameter in pkParameters) {
       singleOuputSinglePKDataframe <- allPKResultsDataframe[ allPKResultsDataframe["QuantityPath"] == output & allPKResultsDataframe["Parameter"] == pkParameter, ]
       quantileResults <- getQuantileIndividualIds(singleOuputSinglePKDataframe, quantileVec)
```
pchelle, line 463: I can't select the loop below, but the lines in loop coming next can be simplified:
instead of `for(i in seq_along...`
You can directly build your data.frame: 
`saResultsByOuptut <- data.frame(Outputs = output, pkParameters = , pkParameter, 
IndividualId = quantileResults$ids, Value = quantileResults$values, ...)`
And then
`sensitivityAnalysesResultsIndexFileDF <- rbind.data.frame(sensitivityAnalysesResultsIndexFileDF,
saResultsByOuptut)`

R/utilities-sensitivity-analysis.R
```diff
@@ -585,14 +616,21 @@ plotPopulationSensitivity <- function(structureSets,
   dFList <- list()
   i <- 0
```
pchelle, line 617: I think `i` is not useful here, it would be better to use the pk parameter name as the list element of dFList line 654

R/utilities-sensitivity-analysis.R
```diff
@@ -618,31 +656,104 @@ plotPopulationSensitivity <- function(structureSets,
...
+      allParamsForThisIndividual <- unique(sensitivityThisOpPk[ (sensitivityThisOpPk$Quantile %in% qu) & (sensitivityThisOpPk$individualId %in% id) & (sensitivityThisOpPk$Population %in% pop), ]$Parameter)
+
+      # create list of the parameters missing for this individual but otherwise shown in this plot for this combination of output and pkParameter
+      missingParameters <- setdiff(allParamsForThisOpPk, allParamsForThisIndividual)
+
+      # loop thru the missing parameters for the current individual
+      for (parNumber in seq_along(missingParameters)) {
+
+        # load the index file of SA results for this individual's population to get the name of the individuals's sensitivity result file
+        indx <- read.csv(saResultIndexFiles[[pop]])
```
pchelle, line 697: this line can be is moved in the previous loop

R/utilities-sensitivity-analysis.R
```diff
@@ -618,31 +656,104 @@ plotPopulationSensitivity <- function(structureSets,
...
+
+        # get the name of the individuals's sensitivity result file
+        saResFileName <- indx[ (indx$Outputs %in% op) & (indx$pkParameters %in% pk) & (indx$Quantile %in% qu), ]$Filename
+
+        # import SA results for individual
+        individualSAResults <- ospsuite::importSensitivityAnalysisResultsFromCSV(
+          simulation = simulationList[[pop]],
+          filePaths = file.path(structureSet$workflowFolder, structureSet$sensitivityAnalysisResultsFolder, saResFileName)
+        )
```
pchelle, line 706: I think you can move these lines in the previous loop and save a few calls of these functions

Yuri05:
In general: sensi plots look much better now. Well done :)
C_max_Organism-PeripheralVenousBlood-Rifampicin-Plasma (Peripheral Venous Blood)

pchelle:
Indeed, great plot.
I have one naive question, are the sensitivities supposed to be within a [-1; 1] range ?

Yuri05:
@pchelle No, can be any value. Is actually an approximation for the derivative `dPKParameter/dModelParameter`

Yuri05:
@abdullahhamadeh I used the code from this PR to create sensitivity plots in a **MeanModel**-Workflow - and got a crash https://github.com/Open-Systems-Pharmacology/OSPSuite.ReportingEngine/issues/214 But I did not try Sensi plots in MeanModel-Workflows for a while, so don't know if the error is new or already there for a while.

abdullahhamadeh:
> @abdullahhamadeh When creating a popWorkflow, popWorkFlow$plotSensitivity$settings is per default of type `SensitivityAnalysisSettings `but must be `SensitivityPlotSettings`

Should be fixed now

R/utilities-sensitivity-analysis.R
abdullahhamadeh: Fixed

R/utilities-sensitivity-analysis.R
abdullahhamadeh: Fixed

R/utilities-sensitivity-analysis.R
abdullahhamadeh: Fixed

R/utilities-sensitivity-analysis.R
abdullahhamadeh: Fixed

R/utilities-sensitivity-analysis.R
abdullahhamadeh: Fixed

abdullahhamadeh:
> @abdullahhamadeh I used the code from this PR to create sensitivity plots in a **MeanModel**-Workflow - and got a crash #214 But I did not try Sensi plots in MeanModel-Workflows for a while, so don't know if the error is new or already there for a while.

Looks like this crash occurs with the current develop branch?

Yuri05:
>Looks like this crash occurs with the current develop branch? 

Pretty sure it was not the develop. Anyway, going to merge now to move forward, will check again then.

PR #215, opened 2020-06-07
213 invalid statement lloq

pchelle:


Yuri05:
@pchelle 
2 Problems with mean model sensitivity in the current PR:
1) If the user does not set any sensitivity parameters explicitly (=> user wants sensitivity for all available parameters), then the error occurs:
[mean2_SensiAll.zip](https://github.com/Open-Systems-Pharmacology/OSPSuite.ReportingEngine/files/4745440/mean2_SensiAll.zip)

```
[1] "08/06/2020 - 12:44:25 : Starting Perform sensitivity analysis task"
[1] "08/06/2020 - 12:44:25 : Run sensitivity for simulation: p1p2p3p4"
Error in self$getTaskResults(structureSet = set, settings = self$settings,  : 
  object 'totalNumberParameters' not found
```
2) If workflow contains 2 or more simulation sets for different simulations, and some sensitivity parameters are available **only in some simulations**, then the error occurs.
Example: [mean2_SensiSelected.zip](https://github.com/Open-Systems-Pharmacology/OSPSuite.ReportingEngine/files/4745464/mean2_SensiSelected.zip)
2 simulations; one has parameters p1,p2,p3, p4 (but not p5); another one has parameters p1,p2,p3,p5 (but not p4). User wants to calculate sensitivity for p1,p2,p3,p4,p5.
```
meanModelWorkflow$meanModelSensitivityAnalysis$settings$variableParameterPaths <- c(
  "Organism|p1", "Organism|p2", "Organism|p3", "Organism|p4", "Organism|p5" )
```
Result:
```
[1] "08/06/2020 - 12:49:27 : Run sensitivity for simulation: p1p2p3p4"
[1] "08/06/2020 - 12:49:27 : Values 'Organism|p1, Organism|p2, Organism|p3, Organism|p4, Organism|p5' are not in included in parent values: 'Organism|Volume, Organism|p1, Organism|p2, Organism|p3'."
Error in logErrorThenStop(messages$errorNotIncluded(values, parentValues)) : 
  Values 'Organism|p1, Organism|p2, Organism|p3, Organism|p4, Organism|p5' are not in included in parent values: 'Organism|Volume, Organism|p1, Organism|p2, Organism|p3'.
```
If I keep only p1, p2, p3 for sensitivity (thus parameters available in both simulations), the same error as in case 1) occurs: 
```
Error in self$getTaskResults(structureSet = set, settings = self$settings,  : 
  object 'totalNumberParameters' not found
```

pchelle:
My bad.
I thought I could simplify some lines of code that actually were used later.
I just removed the last commit which was the cause for this bug.

Yuri05:
Well, for the problem 1.: it does not crash anymore, but also does not produce any sensitivity plots, just an empty report. [Report.zip](https://github.com/Open-Systems-Pharmacology/OSPSuite.ReportingEngine/files/4745867/Report.zip)

pchelle:
I used the same logic as in pk parameter plots: only pk parameters defined in the output objects are plotted by the task.
Here there is no pk parameter in the output objects so the code did not plot anything.
Do we want a different default for sensitivity analysis: all the pk parameters are plotted if nothing is specified in output objects ?

Yuri05:
>I used the same logic as in pk parameter plots: only pk parameters defined in the output objects are plotted by the task.

Yep, my fault, deleted too much from the test workflow. Now I included pk parameters again - and get totally weird sensitivity plots.

[04.zip](https://github.com/Open-Systems-Pharmacology/OSPSuite.ReportingEngine/files/4746644/04.zip)
[Report_with_weird_plots.pdf](https://github.com/Open-Systems-Pharmacology/OSPSuite.ReportingEngine/files/4746658/Report_with_weird_plots.pdf)

Yuri05:
> Do we want a different default for sensitivity analysis: all the pk parameters are plotted if nothing is specified in output objects ?

It's the same as in pop sensitivity analysis: https://github.com/Open-Systems-Pharmacology/OSPSuite.ReportingEngine/issues/171#issuecomment-624548277

PR #260, opened 2020-07-04
253 observed data selection

pchelle:
WIP check if tests pass


pchelle:
Very odd, the tests work on my laptop but not on appveyor... they seem to crash when running the workflow on the Minimodel2

pchelle:
I don't really understand this second fail: 
`Error in curl::curl_fetch_memory(url, handle = handle) : 
  Timeout was reached: [codecov.io] Connection timed out after 10000 milliseconds`
From what I checked there seems to be a connection timeout from codecov because page build might be too long or something, but not quite sure.

pchelle:
For the first fail, the problem is that I wanted to run actual workflows in testthat in order to check their outputs.
However, appveyor needs to install pandoc  if pandoc is used (which is only the case for the creation of the word reports).
There seems to be a way around that but not sure how to implement it: https://github.com/krlmlr/r-appveyor/issues/82

Yuri05:
>There seems to be a way around that but not sure how to implement it: krlmlr/r-appveyor#82

@pchelle Can you add `- ps: cinst pandoc` and `- ps: $env:Path += ";C:\Program Files (x86)\Pandoc\"` to the `install:` section of appveyor.yml and check if it's sufficient?
```
install:
  - ps: Bootstrap
  - git submodule update --init --recursive
  - ps: cinst pandoc
  - ps: $env:Path += ";C:\Program Files (x86)\Pandoc\"
```

Caching (what is described in [your workaround](https://github.com/krlmlr/r-appveyor/issues/82#issuecomment-261695154)) can be added later.

Yuri05:
@pchelle Build still fails.. but why do u think it's becasue of pandoc? In this and also in the previous build all testthat tests were successful. It's the code coverage which fails
```
Rscript -e "covr::codecov(function_exclusions = c('print.*', '\\.onLoad'))"
Error in curl::curl_fetch_memory(url, handle = handle) : 
  Timeout was reached: [codecov.io] Connection timed out after 10006 milliseconds
Calls: <Anonymous> ... request_fetch -> request_fetch.write_memory -> <Anonymous>
```

pchelle:
The PR was failing at 2 levels: 
The first time, it failed because I included workflow tests that needed pandoc. 
Now, that I have included the code you suggested, the tests have passed (I checked on the failure.zip artifact).
However, it fails on that second level which is the codecov timeout. And I have almost no clue why.

msevestre:
@pchelle Codecov timeout is usually a fluke. I can trigger the build again

msevestre:
@pchelle Is it possible to install pandoc silently??
image


This is a log file killer

PR #264, opened 2020-07-08
251 mean model sensitivity plots

abdullahhamadeh:


Yuri05:
@abdullahhamadeh Tried this PR. MEanModel sensitivity plots still look the same with the default settings
https://github.com/Yuri05/RE-Advanced-Examples/blob/master/02_MeanModel_Raltegravir_inclSens/Report/Report.md#11-sensitivity-analysis-for-raltegravir-10-mg-lactose-formulation

Wanted to reduce font size then - but the option has disappeared? What happened?
grafik

abdullahhamadeh:
@Yuri05 I have replaced the `plotFontSize` argument and replaced it with a `plotConfiguration` argument that takes a tlf `PlotConfiguration` object.  I have applied the same change to the pop sensitivity plot for consistency.

To make the font size of the y axis ticks smaller:

`meanModelWorkflow$plotSensitivity$settings = SensitivityPlotSettings$new(totalSensitivityThreshold = 1,
                                                                         maximalParametersPerSensitivityPlot = 3 ,
                                                                         plotConfiguration = tlf::PlotConfiguration$new(ylabel = tlf::Label$new(size = 5)))`

Since the y axis ticks' font size is likely to need scaling down often (because of long paths), I can put a wrapper so that `SensitivityPlotSettings` takes a `plotFontSize` argument again and this is internally translated into the `PlotConfiguration' setting.  Or shall I leave it like this for now?

Yuri05:
>Since the y axis ticks' font size is likely to need scaling down often (because of long paths), I can put a wrapper so that SensitivityPlotSettings takes a plotFontSize argument again and this is internally translated into the `PlotConfiguration' setting. 

Yes, I think it's a good idea.
Please also set the **default** plotFontSize both of MeanModel and Pop-Workflow to 6 (as it was the case for pop workflow before)

abdullahhamadeh:
@Yuri05 you should now be able to set the x and y axis fontsize like so:
```
pwf$plotSensitivity$settings <- SensitivityPlotSettings$new(totalSensitivityThreshold = 1,
                                                            maximalParametersPerSensitivityPlot = 12,
                                                            xAxisFontSize = 10,
                                                            yAxisFontSize = 5)
```

Default for `xAxisFontSize`  and `yAxisFontSize` is 6

Yuri05:
@abdullahhamadeh Nice!

PR #267, opened 2020-07-09
265 pk parameters crash

pchelle:


Yuri05:
@pchelle Can you please change **default** setting of `workflow$createWordReport` to TRUE? Apart from that: works fine.

PR #275, opened 2020-07-21
271 watermak

pchelle:


R/system-info.R
```diff
@@ -1,5 +1,18 @@
 # System Information for Reporting Engine
 
+#' @title isValidatedSystem
+#' @description Check if environment variable is defined and as value \code{'prodrun'}
+#' @return logical assessing if system is validated
+isValidatedSystem <- function() {
+  if (!exists("VALIDATEDR_ENVIRONMENT")) {
```
Yuri05, line 7: @pchelle Will it work? If i try this in R studio with existing environment variables I get false (both on Windows and Linux)
```
> exists("OS")
[1] FALSE
```

Isn't it better to use `Sys.getenv` ? If the environment variable does not exist it returns an empty string.

```
> Sys.getenv("OS")
[1] "Windows_NT"

> Sys.getenv("NON_EXISTING_VARIABLE")
[1] ""
```

pchelle: I think I misunderstood what you meant by "environment variable".
Sys.getenv seems indeed dedicated to environment variables.

PR #277, opened 2020-07-23
248 add test units

pchelle:
WIP


R/utilities-mass-balance.R
```diff
@@ -35,7 +33,7 @@ plotMeanMassBalance <- function(structureSet,
   })]
 
   # User defined coumpound selection
-  selectedCompoundNames <- selectedCompoundNames %||% relevantCompoundNames
+  selectedCompoundNames <- settings$selectedCompoundNames %||% relevantCompoundNames
```
Yuri05, line 36: will this not crash if `settings `is NULL?

pchelle: No, if `settings` is NULL or empty list, `settings$...` will be also NULL

Yuri05: ok. didn't know it works like this in R :)

PR #302, opened 2020-08-11
fixes #301: some bug fixes and changes for tracelib integration (marked by #change) 

### LINKED ISSUE #301: Adjust tracelibUsage
During the second review the following problems in metadata capture were identified:
For the meanModel workflow corresponding code changes are proposed in a pullRequest tracelibIntegration1 (see comments #change ...
 - for the populationModel workflow the corresponding changes need to be applied, too!

1) For capturing metadata of both a) the workflow script/function (written by the user) and b) the runWorkflow method of the workflow-Class two actions need to be used. 
Fortunately there is a way to include the code into the implementation of the ReportEngine, so that the user does not need to call tStart/EndMetadataCapture within his workflow script (at least in the standard case).
One action was implicitly created by the tStartMetadataCapture.
**The second action can easily be created by calling tStartAction directly after tStartMetadataCapture (tEndAction accordingly).**

2) To capture the metadata for a function (e.g. renderReport) the tStartAction/tEndAction pair needs to be called inside the function implementation - not around the call of the function.
**The corresponding calls around renderReport call in runWorkflow method need to  moved into the function renderReport.**

3) As some tasks (e.g. plotTask, simulateTask) have generic methods runTask, which use specific functions passed to the method, the specific function name should be passed as actionNameExtension. 
In loadSimulateTask not the function name but the function body was used [because deparse(substitute(taskFunction)) returns function body for a function variable taskFunction].
**Here the name of the function needs to be determined in the cases.**

4) The relationship between report and plot- and table-files used in the report was not captured in the metadata yet.
This is important e.g. for answering the questions, which reports were created based on a certain model file (e.g. if this contains an error).
In particular not the output files (Report.md, ..., Report.docx) nor the input files (....png, ....csv) were added to the renderReport action. (Instead the reportFileName was added to the runWorkflow).

The following changes are relatively easy:
**Instead of calling tStoreFileMetadata after renderReport in runWorkflow, call tStoreFileMetadata  in renderWordReport - both for md and docx report file.**
**Call tStoreFileMetadata  in addMarkdownToc for the toc file.**

*Capturing the input files for the report is more difficult for multiple reasons:*
a) The usage of table and figure files is stored in the md file not just after creation of all plots, but during the creation of each plot. 
b) The process of inclusion of tables and figures is different:
* for tables is the content of table files inserted in a formatted way into the md file,
* for figures is the path to the plot file inserted as a reference into the md file.
c) Finally the creation of the Word report file including plots and tables seems to be done in a "black box" function knitr::pandoc.

One way to store the input files metadata to the renderReport action is:
**Create temporary files to store the used file names in reset Report,
store the used filenames there within the functions addFigureChunk and addTableChunk,
merge the files with the used filenames in the function mergeMarkdowndFiles, where the md files are merged,
call tStoreFileMetadata for each of the used files in function renderWordReport.**
(Deletion of these helper files need to be done.)

ju-rgen:


R/mean-model-workflow.R
```diff
@@ -91,9 +92,9 @@ MeanModelWorkflow <- R6::R6Class(
       appendices <- appendices[file.exists(appendices)]
       if (length(appendices) > 0) {
         mergeMarkdowndFiles(appendices, self$reportFileName, logFolder = self$workflowFolder)
-        actionToken2 <- re.tStartAction(actionType = "ReportGeneration", actionNameExtension = "runWorkflow")
+        # actionToken2 <- re.tStartAction(actionType = "ReportGeneration", actionNameExtension = "runWorkflow") #change delete
```
Yuri05, line 95: @ju-rgen Here and everywhere where you added _#change delete_Can you please just remove commented lines?

R/mean-model-workflow.R
```diff
@@ -60,6 +60,7 @@ MeanModelWorkflow <- R6::R6Class(
     #' @return All results and plots as a structured output in the workflow folder
     runWorkflow = function() {
       actionToken1 <- re.tStartMetadataCapture(metaDataCapture = TRUE)
+      actionToken2 <- re.tStartAction(actionType = "Run")  #change add
```
Yuri05, line 63: @ju-rgen Here and everywhere: can you please remove the _#change add_ comment?

PR #306, opened 2020-08-19
305 residuals

pchelle:


R/utilities-goodness-of-fit.R
```diff
@@ -902,6 +904,12 @@ getResidualsPlotResults <- function(timeRange, residualsData, metaDataFrame, str
   goodnessOfFitPlots <- list()
   goodnessOfFitCaptions <- list()
 
+  # Residuals can contain Inf values
+  # They need to be removed from the plot and translated to NA in csv output file
+  csvResidualsData <- residualsData
+  csvResidualsData[,"Residuals"] <- removeInf(csvResidualsData[,"Residuals"], logFolder)
```
Yuri05, line 910: so `removeInf()` in fact does not just _remove_ Inf-values but _replaces_ them with NA? 
If so: I would rename the function to `replaceInfWithNA `or something like that.

Yuri05:
@pchelle All residual plots where modified to exclude NA values  - but not the QQ-plot? Is that because QQ-plot can handle NA values?

R/utilities-goodness-of-fit.R
pchelle: Yes, I will rename the function to clarify this

PR #323, opened 2020-09-07
sensitivity analysis vignette

abdullahhamadeh:


pchelle:
get Relative and Absolute paths as task methods is a great idea !

Could you delete the folders/files generated by the vignette and tests examples which you think are not needed.
(For the vignettes, you include the report content as html documents so the source files and folders for the report don't need to be kept to create the vignette)

abdullahhamadeh:
@pchelle should I just use unlink(pwf$workflowFolder, recursive = TRUE) like in your examples?

pchelle:
> 
> 
> @pchelle should I just use unlink(pwf$workflowFolder, recursive = TRUE) like in your examples?

Yep !

For the examples in tests, like ex/ or expar/, I think you can directly delete the output folder or files you don't want.

PR #325, opened 2020-09-10
324 pk parameter crash

pchelle:
WIP (Waiting for Sensitivity Analysis vignette update to pass)
@TWendl Can you confirm if the 2 latest commits should be kept ?


pchelle:
> 
> 
> @pchelle din't not we say that by default we wanted to output NOTHING when they were not specified?
> `ospsuite::allPKParameterNames()` returns the list of ALL potential parameters. They might not exist for the simulation indeed.

I removed it. But if at some point this default behavior needs to be changed, it is centralized there (and consistent between population and mean pk parameters and sensitivity analysis).

msevestre:
I don't understand :) isn't the goal to have the same behaviour everywhere?
using this list is definitely going to lead to some parameters that do not
exist. As long as this is fine, why not

pchelle:
I agree. 
Is there a place to document that so the users can know why we don't want allPKParameters as default behaviour (vignette ?).
By the way, if the build pass, it is not a WIP anymore.

PR #366, opened 2020-11-09
357 documentation sheet

pchelle:


Yuri05:
@pchelle You have changed the excel template WorkflowInput.xlsx.
I am currently changing it as well, so it will conflict. Did you change anything else except renaming the "_Dictionary" sheet?
I would suggest that you tell me all you changes on the excel template and remove it from your PR, I will incorporate then in my version then.

pchelle:
> 
> 
> @pchelle You have changed the excel template WorkflowInput.xlsx.
> I am currently changing it as well, so it will conflict. Did you change anything else except renaming the "_Dictionary" sheet?
> I would suggest that you tell me all you changes on the excel template and remove it from your PR, I will incorporate then in my version then.

From the template in extdata, I think that was the only one. I will remove it then.
I may have changed more in the Excel file of test/dev because I tested case scenarios for checking errors and bugs.

Yuri05:
@pchelle alternatively we can merge AS IS, and I will ovewrite the excel-template after that (after incorporating your changes).
In any case please let me know what you have modified in the template

Yuri05:
>From the template in extdata, I think that was the only one. I will remove it then.

Ok, let's do it like this.

PR #375, opened 2020-11-11
Fixes #364 adjustments for word

### LINKED ISSUE #364: Unclear error message if population file in a PopSimSet is not defined
```
outA <- Output$new(path = "Organism|A|Concentration", displayName = "A", displayUnit = "µg/ml", dataSelection = "TIME < 5", dataDisplayName = "Study A - Data A")
SimulationSet1 <- PopulationSimulationSet$new(simulationSetName = "Set1", referencePopulation = FALSE, populationFile = NULL, simulationFile = "MiniModel2.pkml", outputs = outA)
```

results in the error message
`Error in basename(file) : a character vector argument expected`

The reason is: population file is not set in the call to `PopulationSimulationSet$new`

pchelle:


msevestre:
Merging so that @Yuri05 can try out

PR #380, opened 2020-11-12
Fixes #379

### LINKED ISSUE #379: Crash in plotPKParameters-Task
[bug_PK.zip](https://github.com/Open-Systems-Pharmacology/OSPSuite.ReportingEngine/files/5532392/bug_PK.zip)

```
[1] "12/11/2020 - 17:12:58 : Starting: Plot PK parameters task"
Error in if (is.null(pkParametersTableRef) || all(pkParametersTable$path ==  : 
  missing value where TRUE/FALSE needed
```

abdullahhamadeh:


R/error-checks.R
```diff
@@ -414,14 +414,28 @@ validateSameOutputsBetweenSets <- function(simulationSets, logFolder = NULL) {
...
-      )) {
+
+    if (is.null(pkParametersTableRef)) {
       pkParametersTableRef <- pkParametersTable
       next
     }
+
+    if(all(pkParametersTable$path == pkParametersTableRef$path)){
+      pkParametersTableTest <- NULL
+      for (n in seq_along(pkParametersTable$pkParameter)){
```
pchelle, line 425: Can you replace `n` by `pkParameterIndex` ?

R/error-checks.R
```diff
@@ -414,14 +414,28 @@ validateSameOutputsBetweenSets <- function(simulationSets, logFolder = NULL) {
...
+        if( is.na(pkParametersTable$pkParameter[n]) && is.na(pkParametersTableRef$pkParameter[n]) ) {pkParametersTableTest <- c(pkParametersTableTest,TRUE); next}
+        if( xor(is.na(pkParametersTable$pkParameter[n]),is.na(pkParametersTableRef$pkParameter[n]) ))  {pkParametersTableTest <- c(pkParametersTableTest,FALSE); next}
+        pkParametersTableTest <- c(pkParametersTableTest, pkParametersTable$pkParameter[n] == pkParametersTableRef$pkParameter[n] )
+      }
```
pchelle, line 429: A simpler way, since you loop through the indices of the vectors, would be:
`pkParameterTableTest[pkParameterIndex] <- isIncluded(pkParameterTable$pkParameter[pkParameterIndex], pkParameterTableRef$pkParameter[pkParameterIndex])`
`isIncluded` will account for NAs

R/error-checks.R
```diff
@@ -414,14 +414,28 @@ validateSameOutputsBetweenSets <- function(simulationSets, logFolder = NULL) {
     pkParametersTable <- getPKParametersInSimulationSet(set)
```
pchelle, line 414: Could you update the error message and advise the user they can check the output and pk parameters of simulation sets using the function: `'getPKParametersInSimulationSet'`
Btw, if we decide to not enforce the same outputs, we only need to update this method to throw a warning instead of a stop.

abdullahhamadeh: Done

R/error-checks.R
abdullahhamadeh: Nice - done.

R/error-checks.R
abdullahhamadeh: Done

pchelle (approved):

PR #381, opened 2020-11-16
377 template user task

pchelle:


R/workflow.R
```diff
@@ -15,6 +16,7 @@ Workflow <- R6::R6Class(
     taskNames = NULL,
     reportFileName = NULL,
     createWordReport = NULL,
+    userDefinedTask = NULL,
```
Yuri05, line 19: Should not we provide a **list** of user defined tasks instead? Or what is the reason to restricting userdef tasks to only **one**?

R/utilities-task.R
```diff
@@ -426,3 +426,77 @@ checkTaskInputsExist <- function(task) {
...
+#' @param taskFunctionUsesAllSimulationSets logical if `taskFunction` requires all simulation sets as input, 
+#' `PopulationPlotTask` object is preferred to `PlotTask` object.
+#' @return Updated `Workflow` object
+#' @export
+loadUserDefinedTask <- function(workflow,
+                                taskFunction,
+                                taskName = "userDefinedTask",
+                                active = TRUE,
+                                settings = NULL,
+                                taskFunctionUsesAllSimulationSets = FALSE) {
```
Yuri05, line 448: I wonder if this argument (taskFunctionUsesAllSimulationSets ) is really required.
The decision (**PlotTask** vs. **PopulationPlotTask**) could be done based on the type of the workflow: 
- In case of **MeanModelWorkflow**: Create **PlotTask**
- In case of **PopulationWorkflow**: Create **PopulationPlotTask**

R/workflow.R
pchelle: The reason was during the run of the workflow, I am calling the plot as `self[[taskName]]$run` if userDefinedTask is a list it will crash.
What I can do, is to run the user defined tasks as a loop distinct from the plot tasks, so this slot in the workflow environment can become a list of tasks and won't be limited to only one task.
Would this work better ?

R/utilities-task.R
pchelle: It depends on what kind of task the user intends to do.
As a first implementation of this wrapper, I think your suggestion simplifies a lot, which is better for the users.

R/workflow.R
Yuri05: Yes, I think it would be better

tests/dev/script-template-user-defined-function.R
```diff
@@ -0,0 +1,128 @@
...
+
+  pkRatioTable <- tlf::getPKRatioMeasure(data = pkParameterData, dataMapping = pkRatioMapping)
+  pkRatioTable$Measure <- c("All", "Within 1.5-fold", "Within 2-fold")
+
+  return(list(
+    plots = list("pkratio-AUC" = pkRatioPlot),
+    tables = list("pkratio-AUC" = pkRatioTable),
+    captions = list("pkratio-AUC" = "AUC Ratios across populations")
+  ))
+}
```
Yuri05, line 128: nice!

NAMESPACE
```diff
@@ -55,6 +57,7 @@ export(inactivateWorkflowTasks)
 export(lastPathElement)
 export(loadSimulationWithUpdatedPaths)
 export(loadStudyDesign)
+export(loadUserDefinedTask)
```
Yuri05, line 60: should we call it `addUserDefinedTask `instead of `loadUserDefinedTask`?
When I see `loadUserDefinedTask` I think it would override the (only one) user defined task - which is not the case anymore (now that we have the list of userdef tasks)

Issue #383, opened 2020-11-17
Pop-workflow: Improve default binning for plotPK/plotDemography


Yuri05:
In the attached example 4 populations with 500 individuals each were defined: reference adult population for [30 .. 70] years and 3 children populations for [0 .. 3], [3 .. 6], [6 .. 9] years. 
All the age-dependency plots in PK/Demography show nothing for [0 ..3] and [0..6] populations and only very few bins for [6 ..9].

[01_bin.zip](https://github.com/Open-Systems-Pharmacology/OSPSuite.ReportingEngine/files/5553247/01_bin.zip)
<details><summary><b>Expand to see the plots</b></summary>
<img src="https://user-images.githubusercontent.com/25061876/99386240-73719980-28d2-11eb-99a9-ad52d8940cc8.png">
<img src="https://user-images.githubusercontent.com/25061876/99386521-e549e300-28d2-11eb-8378-f26f7f29c7f8.png">
<img src="https://user-images.githubusercontent.com/25061876/99386549-f266d200-28d2-11eb-8db7-14cb59ec8f39.png">
<img src="https://user-images.githubusercontent.com/25061876/99386756-4376c600-28d3-11eb-87ff-e27063113f38.png">
<img src="https://user-images.githubusercontent.com/25061876/99386802-52f60f00-28d3-11eb-9680-49e0b0e22726.png">
</details>

The default binning algorithm should be improved so that every Range plot shows at least N bins per default.
(not sure what the good choice for N is. 10?)

Yuri05:
@pchelle @abdullahhamadeh Same reason as in #318. I have also tried to set the bins **manually** as described in #318 - but did not manage to get any bins for child populations. Can you tell me hot to set the bins properly for the example attached if I want to have e.g. 10 bins in every range plot?

pchelle:
Currently, binning manually 10 bins per population would give:
__`workflow$plotPKParameters$settings$xParametersBreaks[[StandardPath$Age]] <- unique(c(seq(0, 3, length.out=10), 
                                                                              seq(3, 6, length.out=10), 
                                                                              seq(6, 9, length.out=10), 
                                                                              seq(30, 70, length.out=10)))`__

pchelle:
I went back into the code to see if I could find a solution,
currently the binning happens here: 
[...]

PR #387, opened 2020-11-19
383 improve default binning

pchelle:


R/utilities-calculate-pk-parameters.R
```diff
@@ -285,7 +278,14 @@ plotPopulationPKParameters <- function(structureSets,
           )
 
           for (populationName in populationNames[!populationNames %in% referencePopulationName]) {
-            comparisonData <- aggregatedData[aggregatedData$Population %in% populationName, ]
+            comparisonData <- pkParameterData[pkParameterData$simulationSetName %in% populationName, ]
+            comparisonData <- getDemographyAggregatedData(
+              data = comparisonData,
+              xParameterName = demographyParameter,
+              yParameterName = "Value",
```
Yuri05, line 285: `yParameterName = "Value",` ???

Yuri05 (changes requested):
Range plots with `stairstep = FALSE` look good now.

However with `stairstep = TRUE` (the new default) they don't.
E.g. in the report for the linked issue #383 it looks in the stairstep-Version like this:
grafik

PR #393, opened 2020-11-25
389 missing caption

pchelle:


Yuri05 (changes requested):
Let's don't rename the functions get/setXYParametersForPkParametersPlot.
Otherwise to be consistent other entitities (e.g. PkParameterInfo) should be renamed as well - and I would like to avoid changes in existing RE-interface where possible

pchelle:
> 
> 
> Let's don't rename the functions get/setXYParametersForPkParametersPlot.
> Otherwise to be consistent other entitities (e.g. PkParameterInfo) should be renamed as well - and I would like to avoid changes in existing RE-interface where possible

No problem.

Issue #316, opened 2020-08-28
If a reference simulation is defined in a population workflow, corresponding data (if available) should also be plotted


TWendl:
see comment from sfrechen in https://github.com/Open-Systems-Pharmacology/OSPSuite.ReportingEngine/issues/18

TWendl:
colors for observed data should match colors of actual and references simulation data, i.e. 
concentration time profiles of reference simulation are plotted in green --> observed data beloning to this simulation should be plotted in green symbols, etc.

PR #398, opened 2020-11-30
394 pop time profiles

pchelle:


Yuri05:
@pchelle Looks good. Just one thing: can you write the **name of the reference simulation set** instead of "reference population" in the legend?
grafik

pchelle:
> 
> 
> @pchelle Looks good. Just one thing: can you write the **name of the reference simulation set** instead of "reference population" in the legend?
> grafik

No problem, I am on it
By the way, it partly fixes #316 (reference observed data will appear if dataDisplayName is different between the sets),
however currently there is no match of coloring within the simulations

PR #400, opened 2020-12-03
Fixes #278 display parameters

### LINKED ISSUE #278: Sensitivity plots: parameter display names in RE and in MoBi differ
E.g. take the example from #276. RE parameter paths contain "Organism", MoBi parameter paths do not.

msevestre:
Updating version of ospsuite-r to be 9.1.4 (also with min requireements)

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -85,8 +85,3 @@ branches:
 
 pull_requests:
   do_not_increment_build_number: true
-
-notifications:
-  - provider: Slack
-    incoming_webhook:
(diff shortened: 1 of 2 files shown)


appveyor.yml
```diff
@@ -84,9 +84,4 @@ branches:
     - develop
 
 pull_requests:
-  do_not_increment_build_number: true
-
-notifications:
```
msevestre, line 89: notificaton are now sent twice so I am removing this guy

PR #430, opened 2021-03-03
395 residuals

pchelle:


pchelle:
@Yuri05 
Before merging can you confirm that the linear formula is residuals = (observed-simulated)/simulated
If not, I'll update right away

Yuri05 (approved):

PR #435, opened 2021-03-05
Fixes #434 Add comparison tolerance

### LINKED ISSUE #434: Add comparison tolerance
Still some numeric comparisons are not using a comparison tolerance

Yuri05:


tests/testthat/helper-for-tests.R
```diff
@@ -20,3 +20,7 @@ executeWithTestFile <- function(actionWithFile) {
   actionWithFile(newFile)
   file.remove(newFile)
 }
+
+comparisonTolerance <- function() {
+  1e-4
+}
```
Yuri05, line 26: moved the definition of comparison tolerance here.

pchelle: Good idea.
Would writing `comparisonTolerance <- function(tol = 1e-4) {tol}` be better ?
This way we can also tune to other values if necessary.

msevestre: yes!

Yuri05: done

pchelle (approved):

PR #436, opened 2021-03-08
Fixes #424 decimal numbers in output tables

### LINKED ISSUE #424: Number of decimal places
Especially in PK-Parameter-Tables there are sometimes too many decimal places which results in not so nice looking tables in Word reports. It should be possible to change the default number of decimal places.

grafik

pchelle:
Note that reEnv has been created to centralize RE global variables


R/reportingengine-env.R
```diff
@@ -0,0 +1,33 @@
+# Environment that holds various global variables and settings for ospsuite.reportingengine package,
+# It is not exported and should not be directly manipulated by other packages.
+reEnv <- new.env(parent = emptyenv())
+
+# name of the package. This will be used to retrieve information on the package at run time
+reEnv$packageName <- "ospsuite.reportingengine"
+reEnv$packageVersion <- "1.2"
```
Yuri05, line 7: Why should the package version be stored as part of the global RE environment?
And if it's stored here: why **hardcoded** and not set to `packageVersion("ospsuite.reportingengine")` ?

pchelle: It was also hardcoded in `ospsuite` package when I copied/pasted the global settings and did not even think it through.
However, `packageVersion` can't be used because the code is ran to build the package, so you would get an error by requesting the package version of an non-existing package.
Also, we are currently not using the info of packageName or packageVersion, so not sure this is necessary to be kept.

Yuri05: I didn't see it in the ospsuite package till now and have no idea why it is done like this. Have just created an issue to clarify: https://github.com/Open-Systems-Pharmacology/OSPSuite-R/issues/483 

And if we don't need this in RE: I would remove `reEnv$packageVersion` from the global environment.

Yuri05 (approved):

PR #437, opened 2021-03-11
414 obs data unit

pchelle:


R/error-checks.R
```diff
@@ -287,21 +287,46 @@ logErrorMessage <- function(message, logFolderPath = getwd()) {
 }
 
 validateObservedMetaDataFile <- function(observedMetaDataFile, observedDataFile) {
```
Yuri05, line 289: @pchelle Where is "validateObservedMetaDataFile" called?

pchelle: It is called when creating SimulationSet objects, so within `SimulationSet$new()`

pchelle: https://github.com/Open-Systems-Pharmacology/OSPSuite.ReportingEngine/blob/06dca9dc36a685a1c42b51e8d1eab92bfd3edc45/R/simulation-set.R#L51

Yuri05 (approved):

PR #442, opened 2021-03-22
419 application ranges settings

pchelle:


R/utilities-goodness-of-fit.R
```diff
@@ -738,32 +736,58 @@ plotResidualsQQPlot <- function(data,
...
+  # Initialize output
+  timeRanges <- list(total = list(name = ApplicationRanges$total,
+                                  keep = isTRUE(settings$applicationRanges[[ApplicationRanges$total]]),
+                                  values = NULL),
+                     firstApplication = list(name = ApplicationRanges$firstApplication,
+                                             keep = FALSE,
+                                             values = NULL),
+                     lastApplication = list(name = ApplicationRanges$lastApplication,
+                                            keep = FALSE,
+                                            values = NULL))
```
Yuri05, line 749: @pchelle  I don't unerstand why `keep = FALSE` for the first/last AppRange. Can you explain please?
As **per default**, all 3 ranges should be activated (same as before)

pchelle: Sorry, I should have commented this.
The initialization here defines the default for single administration, for which total = firstApplication = lastApplication.
The chunk from lines 783-787 overwrites this in the case of multiple administrations.

In each case, the default was already defined in `settings` which is a `TaskSettings` object.
When initializing `settings` for plot time profile task, `settings$applicationRanges$total`, `settings$applicationRanges$firstApplication`, and `settings$applicationRanges$lastApplication` are all TRUE.

R/task-settings.R
```diff
@@ -0,0 +1,153 @@
...
+        private$.plotConfigurations <- list(absorptionPlot = NULL)
+      }
+      if (isIncluded(taskName, AllAvailableTasks$plotMassBalance)) {
+        private$.plotConfigurations <- list(
+          timeProfile = NULL,
+          cumulativeTimeProfile = NULL,
+          normalizedTimeProfile = NULL,
+          normalizedCumulativeTimeProfile = NULL,
+          pieChart = NULL
+        )
```
Yuri05, line 34: why is here a `PlotTimeProfilesAndResiduals`-configuration set for the MassBalanceTask?

R/utilities-goodness-of-fit.R
Yuri05: ok, thx for the clarification!

PR #443, opened 2021-03-24
423 descriptor

### LINKED ISSUE #415: Remove simulationName from the definition of a simulation set
Currently a (display) name for the simulation used in a simulation set can be given.
In reports only the name of the simulation set is used in the most cases. `simulationName` is used (additionally to the simulation set name) only in very few places (figure captions of some time profiles and residual plots. s. examples below; XXXXX is the simulation name).

First user experience shows that `simulationName` is not really required and can be removed.

grafik

grafik

### LINKED ISSUE #420: Default axis labels and legend of predicted-vs.-observed plots
“Simulated value” instead of “simulated data”
grafik

### LINKED ISSUE #421: Population workflows: captions for tables (PK parameters) missing

### LINKED ISSUE #422: Population workflows: units for tables (PK parameters) missing
Should it become part of the table (unit column) or part of the table caption (s. #421)?

### LINKED ISSUE #423: New Workflow property: "Descriptor" for the simulation sets
Currently in the report simulation sets are referred to either as "Simulation sets" or "Population" in a population workflow.
However depending on the concrete use case users would like to name simulation sets differently (e.g. "Scenarios").

So it should be possible to (optionally) set this simulation set "Descriptor" in the workflow, which should be used then instead of "Simulation set"/"Population" in the report.

grafik

grafik

pchelle:
PR fixing issues related to captions by creating a central `captions` object working the same was as `messages`
Fixes #423 
Fixes #422 
Fixes #421 
Fixes #420 
Fixes #415 (actually previously fixed by Abdullah but not closed)


R/reportingengine-env.R
```diff
@@ -16,6 +16,7 @@ reEnv$binUsingQuantiles <- TRUE
 reEnv$defaultStairstep <- TRUE
 
 reEnv$defaultWatermarkMessage <- "preliminary analysis"
+reEnv$defaultDescriptor <- ""
```
Yuri05, line 19: let's name it **defaultSimulationSetDescriptor**

R/utilities-calculate-pk-parameters.R
```diff
@@ -152,11 +159,13 @@ plotPopulationPKParameters <- function(structureSets,
   yParameters <- yParameters %||% structureSets[[1]]$simulationSet$outputs
   # Get first simulation, in case mol weight is needed
   simulation <- loadSimulationWithUpdatedPaths(structureSets[[1]]$simulationSet)
+  setDescriptor <- structureSets[[1]]$simulationSetDescriptor
```
Yuri05, line 162: let's name it **simulationSetDescriptor**

R/utilities-demography.R
```diff
@@ -26,58 +26,60 @@ plotDemographyParameters <- function(structureSets,
 
   demographyPlots <- list()
   demographyCaptions <- list()
-  captionSimulationNames <- paste0(as.character(sapply(structureSets, function(set) {
-    set$simulationSet$simulationSetName
-  })), collapse = ", ")
+  setDescriptor <- structureSets[[1]]$simulationSetDescriptor
```
Yuri05, line 29: same

R/utilities-writing-report.R
```diff
@@ -369,3 +369,25 @@ addMarkdownToc <- function(tocContent, fileName, logFolder = getwd()) {
...
+  return(invisible())
+}
+
+#' @title getSimulationDescriptor
+#' @description Get workflow simulation set descriptor
+#' @param workflow A `Workflow` object
+#' @return character describing simulation sets 
+getSimulationDescriptor <- function(workflow){
+  validateIsOfType(workflow, "Workflow")
+  retrun(workflow$getSimulationDescriptor())
```
Yuri05, line 392: `return `instead of `retrun`
seems there are no unit tests for the function so far 😉

Yuri05 (changes requested):
s. commenst

R/utilities-writing-report.R
pchelle: Good catch ! Thanks

Yuri05 (approved):

PR #460, opened 2021-04-13
Fixes #419 Application Ranges defined from SimulationSet objects

### LINKED ISSUE #419: Allow switching application ranges off
For the time profile plots in multiple administration simulation 3 time ranges are plotted: full range, first application range, last application range.
It should be possible to deactivate some of them (e.g. plot last application range only)

pchelle:


R/simulation-set.R
```diff
@@ -25,22 +28,29 @@ SimulationSet <- R6::R6Class(
...
+    #' Providing directly a logical value will associate the value to all application ranges.
     #' @return A new `SimulationSet` object
     initialize = function(simulationSetName,
                               simulationFile,
                               outputs = NULL,
                               observedDataFile = NULL,
                               observedMetaDataFile = NULL,
-                              timeUnit = "h") {
+                              timeUnit = "h",
+                              applicationRanges = TRUE) {
```
Yuri05, line 41: `applicationRanges = TRUE` ??
I expected the structure with 3 fields (first, last, total) (each of them TRUE per default)?

pchelle: The input also accepts a named list with firstApplication, lastApplication and total.
If the user input `TRUE`/`FALSE` directly all 3 values become true/false.
If it is better for users, I can replace `applicationRanges` by 3 separate inputs  firstApplication, lastApplication and total.

Yuri05: I would in fact suggest to provide the 3-list only.
In the current setup, TRUE is the default and can be omitted anyway, FALSE does not really make sense (why would you define time profile task without any time profiles) - thus if users want to override the setting, they will always provide a 3-list.

PR #467, opened 2021-04-23
459 sensitivity line breaks

pchelle:


R/reportingengine-env.R
```diff
@@ -18,6 +18,9 @@ reEnv$defaultStairstep <- TRUE
 reEnv$defaultWatermarkMessage <- "preliminary analysis"
 reEnv$defaultSimulationSetDescriptor <- ""
 
+reEnv$maximalParametersPerSensitivityPlot <- 15
```
Yuri05, line 21: @pchelle Didn't we say 25? I suggest to set it to 25 per default

pchelle: We definitely did. Sorry about the typo.
I am going to fix that.

PR #481, opened 2021-05-15
WIP select plot outputs for qualification

abdullahhamadeh:


R/utilities-goodness-of-fit.R
```diff
@@ -33,8 +33,19 @@ plotMeanGoodnessOfFit <- function(structureSet,
...
 
+
+  outputSelections <- structureSet$simulationSet$outputs
+  if (!is.null(settings$outputSelections)) {
+    outputSelections <- NULL
+    for (output in structureSet$simulationSet$outputs) {
+      if (output$path %in% settings$outputSelections) {
+        outputSelections <- c(outputSelections, output)
+      }
+    }
```
pchelle, line 44: Maybe, if you want to save some if/for loops, you can try to use directly the path names with sapply
```
availableOutputs <- sapply(structureSet$simulationSet$outputs, function(output){output$path})
selectedOuptuts <- availableOutputs %in% settings$outputSelections
outputSelections <- structureSet$simulationSet$outputs[selectedOuptuts]
```

R/utilities-qualification.R
```diff
@@ -0,0 +1,235 @@
...
+  return(appendices)
+}
+
+
+
+#' @title getOutputsTimeProfile
+#' @description Get a vector of output paths that are to be used in a time profile plot
+#' @param plot is a descriptor of the plot that is read from the `ConfigurationPlan` that is generated by the configuration plan json
+#' @return A vector of output paths that are to be used in the time profile plot descriptor `plot`
+getOutputsTimeProfile <- function(plot) {
```
pchelle, line 163: I think the function name should be more explicit to clarify that "plot" is an object from the config plan,
something like: getOutputsFromTimeProfileConfiguration

R/utilities-qualification.R
```diff
@@ -0,0 +1,235 @@
...
+  }
+  return(unique(paths))
+}
+
+
+#' @title getOutputsGOFMergedPlotsPlot
+#' @description Get a vector of output paths that are to be used in a GOF merged plot
+#' @param plot is a descriptor of the plot that is read from the `ConfigurationPlan` that is generated by the configuration plan json
+#' @return A vector of output paths that are to be used in the GOF merged plot descriptor `plot`
+getOutputsGOFMergedPlotsPlot <- function(plot) {
```
pchelle, line 183: same here

R/utilities-qualification.R
```diff
@@ -0,0 +1,235 @@
...
+  validateIsIncluded(values = "Groups", parentValues = names(plot), nullAllowed = TRUE)
+  paths <- NULL
+  for (group in plot$Groups) {
+    validateIsIncluded(values = "OutputMappings", parentValues = names(group), nullAllowed = TRUE)
+    for (outputMapping in group$OutputMappings) {
+      validateIsIncluded(values = "Output", parentValues = names(outputMapping), nullAllowed = TRUE)
+      validateIsString(object = outputMapping$Output)
+      print(outputMapping$Output)
```
pchelle, line 191: I guess there was some debugging here :)

R/utilities-qualification.R
```diff
@@ -0,0 +1,235 @@
...
+  validateIsString(object = plotType)
+  schedule <- NULL
+  plotNumber <- 0
+  for (plt in pltList) {
+    plotNumber <- plotNumber + 1
+    schedule <- rbind.data.frame(schedule, data.frame(plotType = plotType, plotNumber = plotNumber, sectionId = plt$SectionId))
+  }
+  return(schedule)
+}
+
```
pchelle, line 213: @abdullahhamadeh , can you clarify the purpose of this function with a practical example ?
I am not sure why the plotNumber may be needed

And just to be consistent with the nomenclature you used, can you rename pltList into something like plotList or listOfPlots

msevestre:
@abdullahhamadeh #479 was merged. Can you rebase/merge your changes on develop?

R/utilities-goodness-of-fit.R
abdullahhamadeh: Changed

R/utilities-qualification.R
abdullahhamadeh: Changed

R/utilities-qualification.R
abdullahhamadeh: Changed

R/utilities-qualification.R
abdullahhamadeh: Removed

R/utilities-qualification.R
abdullahhamadeh: @pchelle This function is still WIP.  At one point I thought it would be good to have a dataframe that maps each plot a simulation set, and I started adding fields in the `schedule` dataframe that indicate 'project' and 'simulation'.  But there will be plots that are associated with more than one simulation set (see eg GOF plots and comparison plots in config plan), so the `schedule` dataframe may not be adequate on its own.

msevestre (commented):
Looks good to me

@abdullahhamadeh @pchelle 
One comment: Should we try to move the strings into their own files for error messages  or description? That way, it is easy to review the syntax all at once instead of scattered everywhere in the code. Let me know your thoughts

PR #494, opened 2021-06-22
Functions to build qualification GOF plots

abdullahhamadeh:


R/utilities-qualification.R
```diff
@@ -202,3 +202,65 @@ getOutputsFromGOFMergedPlotsConfiguration <- function(plot) {
...
+    unit = ifelse(test = is.na(x = splitVariableUnitString[2]), yes = "", no = splitVariableUnitString[2])
+  ))
+}
+
+
+#' @title parseObservationsDataFrame
+#' @description Function to read the variable names and units in the first two columns of an observations data frame used for qualification. First column stores timepoints.  Second column stores measurements of a quantity corresponding to timepoints in first column.
+#' @param observationsDataFram
+#' @return A named list with `time` and `output` fields.  Each field contains a list that is output by `separateVariableFromUnit` with fields that store the variable name and the unit.
+parseObservationsDataFrame <- function(observationsDataFrame) {
```
pchelle, line 226: @abdullahhamadeh , there are some cases in which there is a 3rd column for uncertainty.
Should we include a 3rd field for uncertainty in that list ?

R/utilities-qualification.R
```diff
@@ -202,3 +202,65 @@ getOutputsFromGOFMergedPlotsConfiguration <- function(plot) {
...
+
+
+#' @title massMoleConversion
+#' @description Function to map `Concentration (mass)` or `Mass` dimensions to `Concentration (molar)` and `Amount` respectively.
+#' @param dimension is string that is from among the valid OSP dimensions
+#' @return If `dimension` is `Concentration (mass)` or `Mass`, then return `Concentration (molar)` or `Amount` respectively, otherwise return `dimension`.
+massMoleConversion <- function(dimension) {
+  massMoleConversionList <- list()
+  massMoleConversionList[[ospDimensions$Mass]] <- ospDimensions$Amount
+  massMoleConversionList[[ospDimensions$`Concentration (mass)`]] <- ospDimensions$`Concentration (molar)`
```
pchelle, line 243: In case, to prevent bugs of namespace and such, could you use `ospsuite::ospDimensions` ?

R/utilities-qualification-gof.R
```diff
@@ -0,0 +1,409 @@
...
+
+    plotGOFMetadata$axesSettings <- lapply(plotGOFMetadata$plotTypes, function(pltType) {
+      getAxesSettings(configurationPlan$plots$AxesSettings[[gofPlotAxesSettings[[pltType]]]])
+    })
+    names(plotGOFMetadata$axesSettings) <- plotGOFMetadata$plotTypes
+
+    plotGOFMetadata$groups <- list()
+
+
+    for (group in seq_along(gofPlotConfiguration$Groups)) {
```
pchelle, line 26: If we do not need to use the index, I suggest we use directly the loop on object itself:
```
for(group in gofPlotConfiguration$Groups){
caption <- group$Caption
symbol <- group$Symbol
outputMappings <- group$OutputMappings
...
}
```

R/utilities-qualification-gof.R
```diff
@@ -0,0 +1,409 @@
...
+        outputMapping <- outputMappings[[omap]]
+
+        projectName <- outputMapping$Project
+        simulationName <- outputMapping$Simulation
+        outputPath <- outputMapping$Output
+        color <- outputMapping$Color
+
+        observedDataPathInSimulation <- outputMapping$Output
+        observedDataSet <- outputMapping$ObservedData
+        observedDataSetFilePath <- configurationPlan$observedDataSets[configurationPlan$observedDataSets$id == outputMapping$ObservedData, ]$path
```
pchelle, line 47: I think there is a method in configuration plan for that: 
`configurationPlan$getObservedDataPath(id = outputMapping$ObservedData)`

R/utilities-qualification-gof.R
```diff
@@ -0,0 +1,409 @@
...
+      caption <- gofPlotGroup$Caption
+      symbol <- gofPlotConfiguration$Groups[[group]]$Symbol
+      outputMappings <- gofPlotGroup$OutputMappings
+
+      plotGOFMetadata$groups[[caption]] <- list()
+      plotGOFMetadata$groups[[caption]]$outputMappings <- list()
+      plotGOFMetadata$groups[[caption]]$symbol <- symbol
+
+
+      for (omap in seq_along(outputMappings)) {
```
pchelle, line 37: Same here.
Besides, if you end up using seq_along, could you call the variable `mappingIndex` instead of `omap`, which helps identifying that the variable should be an integer (sorry if I am picky about the naming)

R/utilities-qualification-gof.R
```diff
@@ -0,0 +1,409 @@
...
+          simulation = simulationName
+        )
+
+        simulationResultsFile <- configurationPlan$getSimulationResultsPath(
+          project = projectName,
+          simulation = simulationName
+        )
+
+        simulation <- loadSimulation(simulationFile, loadFromCache = TRUE)
+        simulationResults <- importResultsFromCSV(simulation = simulation, filePaths = simulationResultsFile)
```
pchelle, line 60: In case add `ospsuite::` for namespace

R/utilities-qualification-gof.R
```diff
@@ -0,0 +1,409 @@
...
+
+
+        # Setup simulations dataframe
+        simulatedDataStandardized <- data.frame(
+          Time = simulationResults$timeValues,
+          Concentration = simulationResults$getValuesByPath(path = outputPath, individualIds = 0)
+        )
+
+        # Setup observations dataframe
+        observedDataFileData <- read.csv(file.path(inputFolder, observedDataSetFilePath), check.names = FALSE, fileEncoding = "UTF-8-BOM")
```
pchelle, line 81: Why not using the RE function `readObservedDataFile` whose default encoding is set as UTF-8 ?

R/utilities-qualification-gof.R
```diff
@@ -0,0 +1,409 @@
...
+
+        outputs <- lapply(simulation$outputSelections$allOutputs, function(output) {
+          Output$new(output$path)
+        })
+        names(outputs) <- lapply(simulation$outputSelections$allOutputs, function(output) {
+          output$path
+        })
+        output <- outputs[[outputPath]]
+        molWeight <- simulation$molWeightFor(outputPath)
+        simulationDimension <- getQuantity(path = outputPath, container = simulation)$dimension
```
pchelle, line 70: add `ospsuite::` in case

R/utilities-qualification-gof.R
```diff
@@ -0,0 +1,409 @@
...
+        observedDataFileData <- read.csv(file.path(inputFolder, observedDataSetFilePath), check.names = FALSE, fileEncoding = "UTF-8-BOM")
+        observedDataFileMetaData <- parseObservationsDataFrame(observedDataFileData)
+        # If simulation in c("Amount","'"Concentration (molar)") and observations in c("Mass","'"Concentration (mass)"), convert observations to c("Amount","'"Concentration (molar)")
+        if (simulationDimension %in% c(ospDimensions$Amount, ospDimensions$`Concentration (molar)`)) {
+          observationsDimension <- massMoleConversion(ospsuite::getDimensionForUnit(observedDataFileMetaData$output$unit))
+        }
+        # Verify that simulations and observations have same dimensions
+        validateIsIncluded(values = observationsDimension, parentValues = simulationDimension, nullAllowed = FALSE)
+
+        observedDataStandardized <- observedDataFileData[, c(1, 2)]
```
pchelle, line 90: @Yuri05 ,
Just wondering, at some point, would there be an interest to include the uncertainty around observed data in the GOF merged plots ?

R/utilities-qualification-gof.R
```diff
@@ -0,0 +1,409 @@
...
+        )
+        observedDataStandardized$Concentration <- ospsuite::toBaseUnit(
+          quantityOrDimension = observationsDimension,
+          values = observedDataStandardized$Concentration,
+          unit = observedDataFileMetaData$output$unit,
+          molWeight = molWeight
+        )
+
+
+        commonTimePoints <- intersect(observedDataStandardized$Time, simulatedDataStandardized$Time)
```
pchelle, line 105: Does `interest` capture all the data ? 
In residuals of RE, I am assuming that the time points may not perfectly match and that the simulation points are dense enough to be close to the observed data. Consequently, I am calculating the closest simulation time points to the observed time points in function `getResiduals` in utilities-goodness-of-fit.R
If it does I could replace the current method. Otherwise, you can leverage it if you want.

abdullahhamadeh: I think the only robust way to deal with this is to read the time points in the observed data set and explicitly add them to the simulation before simulating.

eg
```
timepointsFromObsDataCSV <- c(0, 30, 60, 120, 240, 480, 960, 1440)
simulation$outputSchema$addTimePoints(timepointsFromObsDataCSV)
```

Then we can use `intersect` cleanly.

This will involve going through the dataset/simulation pairs and identifying the timepoints when setting up the workflow.

R/utilities-qualification-gof.R
abdullahhamadeh: Changed

R/utilities-qualification-gof.R
abdullahhamadeh: Done

R/utilities-qualification-gof.R
abdullahhamadeh: Changed to `mappingIndex` since the actual index is useful for later on

R/utilities-qualification-gof.R
abdullahhamadeh: Changed

R/utilities-qualification-gof.R
abdullahhamadeh: Done

R/utilities-qualification.R
abdullahhamadeh: Added

R/utilities-qualification.R
abdullahhamadeh: Done

R/utilities-qualification-gof.R
abdullahhamadeh: Added

PR #500, opened 2021-07-08
Observed Data in time profile plots are converted to right unit

pchelle:
Besides, uncertainty is also plotted as error bars: 
if no unit, uncertainty is a fraction and plotted as geometric error,
if there is a unit, it is converted to display unit and plotted as arithmetic error


R/utilities-time-profile.R
```diff
@@ -27,16 +27,33 @@ plotQualificationTimeProfiles <- function(configurationPlan,
     simulation <- loadSimulation(simulationFile)
     simulationResults <- importResultsFromCSV(simulation, simulationResultsFile)
 
+    # Get mol weight for observed data as first molecule path
+    moleculePath <- utils::head(ospsuite::getAllMoleculePathsIn(simulation), 1)
+    molWeightForObservedData <- simulation$molWeightFor(moleculePath)
```
Yuri05, line 32: Why "first molecule path" and not by the molecule name???
The order of molecules in simulation is random

R/utilities-configuration-plan.R
```diff
@@ -117,19 +119,58 @@ getCurvePropertiesForTimeProfiles <- function(configurationPlanCurve,
...
+    )
+    outputValues <- ospsuite::toUnit(
+      quantityOrDimension = ospsuite::getDimensionForUnit(observedResults$metaData$output$unit),
+      values = observedResults$data[, 2],
+      targetUnit = axesProperties$y$unit,
+      sourceUnit = observedResults$metaData$output$unit,
+      molWeight = molWeight
+    )
+
+    outputUncertainty <- NULL
```
Yuri05, line 138: here and everywhere: let's use the same terminology as in PK-Sim/MoBi/etc. and call it "Error" instead of "Uncertainty"

R/utilities-time-profile.R
pchelle: The issue is that the path defined by observed data curve does not necessarily match with the molecule name of the simulation. 
For instance, in the first plot of minimal example, the molecule should be "Midazolam" however, the observed data path is "Midazolam 600mg SD|ObservedData|Peripheral Venous Blood|Plasma|Rifampicin|Conc" which would lead to an empty molecule path because "Rifampicin" is not in the simulation. This create an NA value for molWeight and crashes the workflow because `toUnit` cannot convert the mg/l from the source data into umol/l for the display plot.

Consequently, the strategy I adopted was to use and update the molWeight from the simulated data because it uses an actual quantity path which should be for the same molecule.
But I wanted to prevent the loop to break if there is no simulated data to plot, so I started with the first molecule path I can find in the simulation.

I know this is not optimal, what would you suggest to improve this ?

Yuri05: Who created this funny minimal example, lol

I would say:
* In the first place, the qualification runner should export a map "Observed dataset name <=> Molweight", which would be read by the reporting engine then
* The reporting engine should estimate MW like this:
      1. Try to get the MW from this map
      2. If it's not there: try to get the MW from the simulation using the **compound name** from the observed data set
      3. If it's not there: set MW to NaN (don't throw an exception)

Later (e.g. when plotting an observed data) exception will be thrown only if MW is NaN AND it's really required (e.g. in cases where observed data is stored in mass units but must be plotted in molar units).

So for now I would suggest to implement steps 2 and 3.

Makes sense?

@msevestre Any thoughts on this?

pchelle: @Yuri05 
In the case of the exception, how the error should be handled:
1- Throw an actual error and stop the workflow
2- Throw a warning and use the observed data as is (e.g. message can be something like : `Molecular weight required for observed data Id 'Midazolam 600mg SD', but not found. Observed data values were used as is`)
3- Throw a warning and don't use the data (e.g. message can be something like : `Molecular weight required for observed data Id 'Midazolam 600mg SD', but not found. Observed data not used in time profile plot`)

Yuri05: My suggestion is:
* If MW cannot be estimated: set it to NaN and show a warning; no exception
* Exception should be thrown only during unit conversion if **MW is required** but is NaN. 
  * E.g. if observed data is given in molar units and must be plotted in molar units as well: MW is not required for the unit conversion, thus even if it's NaN it's not a problem

pchelle: Sorry by exception, I meant the case when MW is NaN and it is required for display unit.

Yuri05: >by exception, I meant the case when MW is NaN and it is required for display unit. 

Ok, in this case I would throw an exception and stop the workflow. In such a case there is not enough information provided so the workflow cannot continue. User must either provide a MW then or adjust a qualification plan and exclude this observed data set from the report.

R/configuration-plan.R
```diff
@@ -231,6 +240,9 @@ ConfigurationPlan <- R6::R6Class(
         path = sapply(value, function(mapping) {
           mapping$Path
         }),
+        molWeight = sapply(value, function(mapping) {
+          mapping$MolWeight %||% mapping$MW %||% NA
```
pchelle, line 244: I don't know yet how the Qualification Runner will name such a field, but if necessary I can update here.
Currently, this reads as look for a field named "MolWeight", then "MW" if nothing molWeight is NA

R/utilities-configuration-plan.R
```diff
@@ -117,12 +117,69 @@ getCurvePropertiesForTimeProfiles <- function(configurationPlanCurve,
...
+        quantityOrDimension = ospsuite::getDimensionForUnit(observedResults$metaData$output$unit),
+        values = observedResults$data[, 2],
+        targetUnit = axesProperties$y$unit,
+        sourceUnit = observedResults$metaData$output$unit,
+        molWeight = molWeight
+      )
+    },
+    error = function(e) {
+      NULL
+    }
```
pchelle, line 143: As written in the commented line, toUnit crashes **only if MW is NA and required**.
I used `tryCatch` to prevent the long unnecessary rClr message about toUnit error replaced by the following error message : "Molecular weight not found but required for observed data Id '<obs data id>' in Time Profile plot."

R/configuration-plan.R
```diff
@@ -132,9 +132,18 @@ ConfigurationPlan <- R6::R6Class(
...
+    
+    #' @description Get molecular weight of observed data corresponding to a specific observedDataSet Id
+    #' @param id observedDataSet identifier
+    #' @return The observed data file path corresponding to the id in the configuration plan field `observedDataSet`
+    getMolWeightForObservedData = function(id) {
+      validateIsIncluded(id, private$.observedDataSets$id, groupName = "'id' variable of observedDataSets")
+      selectedId <- private$.observedDataSets$id %in% id
+      # In case of duplicate observed data, use first
+      return(utils::head(private$.observedDataSets$molWeight[selectedId], 1))
     },
```
Yuri05, line 147: don't understand the implementation

R/utilities-configuration-plan.R
```diff
@@ -117,12 +117,69 @@ getCurvePropertiesForTimeProfiles <- function(configurationPlanCurve,
 
   # Observed Data
   if (pathArray[2] %in% "ObservedData") {
-    observedData <- getObservedDataFromConfigurationPlan(pathArray[1], configurationPlan, logFolder)
+    observedResults <- getObservedDataFromConfigurationPlan(pathArray[1], configurationPlan, logFolder)
+    molWeight <- getMolWeightForObservedData(pathArray, configurationPlan, simulation)
```
Yuri05, line 121: @pchelle Please add some comments with e.g. an example how such a path of observed data or of simulated data looks like in the configuration plan. Otherwise it's not easy to understand, what is `pathArray[1]`, `pathArray[2]`, etc.

R/utilities-configuration-plan.R
```diff
@@ -117,12 +117,69 @@ getCurvePropertiesForTimeProfiles <- function(configurationPlanCurve,
 
   # Observed Data
   if (pathArray[2] %in% "ObservedData") {
-    observedData <- getObservedDataFromConfigurationPlan(pathArray[1], configurationPlan, logFolder)
+    observedResults <- getObservedDataFromConfigurationPlan(pathArray[1], configurationPlan, logFolder)
+    molWeight <- getMolWeightForObservedData(pathArray, configurationPlan, simulation)
+
+    time <- ospsuite::toUnit(
+      quantityOrDimension = "Time",
+      values = as.numeric(observedResults$data[, 1]),
```
Yuri05, line 125: please add comments explaining how `data `is structured. (e.g. that `data[, 1] `is the times vector etc.)

R/configuration-plan.R
pchelle: `ConfigurationPlan` objects  include a field named `observedDataSets` that translated the same field in the json file of the configuration plan. This field is a data.frame with `id`, `path`, `type` and now `molWeight` variables
So to get a path or a specific observed dataset `id`, the method 1) checks that `id` is included in the config plan; 2) select the row corresponding to the id (in observed data set, there were cases of duplicate ids and paths in which case they are identical but only one value needs to be returned); 3) return the `path` or `molWeight` for that selected row.

R/utilities-configuration-plan.R
```diff
@@ -117,12 +117,69 @@ getCurvePropertiesForTimeProfiles <- function(configurationPlanCurve,
...
+        ),
+        logFolderPath = logFolder
+      )
+    }
+
+    outputError <- NULL
+    if (!isOfLength(observedResults$metaData$error, 0)) {
+      # No unit means that error is geometric
+      outputError$ymin <- outputValues / observedResults$data[, 3]
+      outputError$ymax <- outputValues * observedResults$data[, 3]
```
Yuri05, line 158: it's elementwise division, right? what will happen if some entries in the error-vector(`observedResults$data[, 3]`) are NaN?

R/utilities-configuration-plan.R
```diff
@@ -117,12 +117,69 @@ getCurvePropertiesForTimeProfiles <- function(configurationPlanCurve,
...
+
+      if (!isIncluded(observedResults$metaData$error$unit, "")) {
+        outputError$ymin <- outputValues - ospsuite::toUnit(
+          ospsuite::getDimensionForUnit(observedResults$metaData$error$unit),
```
Yuri05, line 162: same question as above

R/utilities-configuration-plan.R
pchelle: Indeed, unlike Matlab elementwise multiplication and division is the default in R (%*% and %/% are the operators for matrix calculation)
Dividing or multiplying by NA elements leads to the NA elements in the same position in the result vector.

R/utilities-configuration-plan.R
pchelle: Same as above, adding or removing a vector with NA elements leads to the NA elements in the same position in the result vector.

R/utilities-configuration-plan.R
Yuri05: Actually it would be better to refactor and define some additional functions like:
isObservedData(pathArray)
observedDataId(pathArray)
etc.
In those functions I would comment how the pathArray looks like; here in the "main" function pathArray would not appear at all.

pchelle: Yes, I'll add both as comment of the code and in the documentation of the `getObservedDataFromConfigurationPlan` fonction

R/utilities-configuration-plan.R
```diff
@@ -214,3 +272,28 @@ tlfScale <- function(configurationScale) {
...
   ConfigurationScales[[tolower(configurationScale)]]
 }
+
+#' @title get
+#' @description Translate value of `Scaling` property from configuration plan into tlf `scale`
+#' @param pathArray Configuration plan path as an array for observed data
+#' @param configurationPlan configuration plan object
+#' @param simulation simulation object
+#' @return molecular weight in base unit or NA if not found
+getMolWeightForObservedData <- function(pathArray, configurationPlan, simulation) {
```
Yuri05, line 282: please refactor and pass observedDataId, compoundName etc. as arguments instead of pathArray, e.g.

`getMolWeightForObservedData <- function(observedDataId, compoundName, configurationPlan, simulation) `

R/utilities-configuration-plan.R
```diff
@@ -214,3 +272,28 @@ tlfScale <- function(configurationScale) {
...
+  if (!is.na(molWeight)) {
+    return(molWeight)
+  }
+  # pathArray is assumed to include the compound name as before the last value
+  compoundName <- utils::head(utils::tail(pathArray, 2), 1)
+  # In the current version, getAllMoleculePathsIn is faster and lighter than getAllMoleculesMatching
+  # since only a path name for the molecule is necessary
+  allMoleculePaths <- ospsuite::getAllMoleculePathsIn(simulation)
+  pathForCompoundName <- utils::head(allMoleculePaths[grepl(compoundName, allMoleculePaths)], 1)
+  # When compound is not found in simulation, return NA
```
Yuri05, line 294: not sure if it's the optimal way. 
Created an issue for that: #507 
For now we can take this implementation but we should check if there is a better way for doing this in the ospsuite

R/utilities-configuration-plan.R
Yuri05: and `ospsuite::toUnit` can also handle NaNs and will return NaN output value for NaN-input?

pchelle: I just checked, yes, that is the case

Issue #510, opened 2021-07-22
Default plot configuration from Configuration Plan


pchelle:
Default plot configuration as defined in `Theme` objects should be read from `ConfigurationPlan`

pchelle:
From the current examples, the available configuration plan fields for the default properties are:
```
"PlotSettings": {
      "ChartWidth": 500,
      "ChartHeight": 400,
      "Fonts": {
        "AxisSize": 11,
        "LegendSize": 9,
        "OriginSize": 9,
        "FontFamilyName": "Arial",
        "WatermarkSize": 40
      }
```
Plot-specific properties are defined slightly differently: 
```
"Plot": {...,
"FontAndSize": {
            "ChartWidth": 600,
            "ChartHeight": 400,
            "Fonts": {
              "AxisSize": 10,
              "LegendSize": 8,
              "TitleSize": 16,
              "DescriptionSize": 12,
              "OriginSize": 8,
              "FontFamilyName": "Microsoft Sans Serif",
              "WatermarkSize": 32
            }
          },
          "Settings": {
            "SideMarginsEnabled": true,
            "LegendPosition": "None",
            "BackColor": "#FFFFFF",
            "DiagramBackColor": "#FFFFFF"
          }
}
```

pchelle:
Currently, I am trying to pass on all default plot settings arguments in a `Theme` object, 
then, during the task I create `PlotConfiguration` objects which take their default from the `Theme` and update the fields that are defined in a plot-specific way.

pchelle:
@Yuri05 , I have a few questions regarding that implementation:
1- Do we expect more fields than the ones above ? If so, were can I find them ?
2- Can you confirm that the plot size (ChartWidth and ChartHeight) are in pixels ? If so can we add a ChartUnits field ? (the reason is that ggplot2 v3.3.0 do not account for pixels even if later versions do, so I need to convert from pixel to inches first)
3- More specific to the fields described in the example above:
   - Does AxisSize accounts for ticklabels only or both xlabel/ylabel and ticklabels ?
   - Does DescriptionSize correspond to subtitle ?
   - Settings include LegendPosition = "None", does this mean I should hide legend in such plots ?
   - I can update the background of the plot and of the area besides the plot. Does BackColor field correspond to both or just the area around the plot ?

PR #515, opened 2021-07-31
512 config sections

### LINKED ISSUE #512: Make title and content optional in sections
if content is defined, use it and do not use the title property
If content is not defined, use title

if neither content or title is defined, the json is invalid (@Yuri05 to check if it's possible)

### LINKED ISSUE #513: Update table of content levels
Right now it is set to 3 but it needs to be 6
method
`numberSections`

### LINKED ISSUE #514: Use title for Time Profile
@pchelle @Yuri05 We discussed that this could be exported by pksim but it can't. OItherwise, loading the snapshot would add a title all the time
image

So the logic should be
1/If title is define, use it
2/otherwise, use the name

pchelle:
Fixes #512
Fixes #513
Fixes #514
I kept some modifications that are intended to support the fix of #510, however it is not entirely fixed yet


NAMESPACE
```diff
@@ -110,10 +109,10 @@ export(runSensitivity)
 export(saveTaskResults)
 export(setDefaultBins)
 export(setDefaultNumericFormat)
+export(setDefaultPlotFormat)
 export(setDefaultStairstep)
 export(setDefaultTheme)
 export(setDefaultThemeFromJson)
-export(setPlotFormat)
```
Yuri05, line 116: @pchelle We should try to keep the changes of the **public** RE interface as small as possible.
If there are good reasons for refactoring - sure, we should make changes. 
But here I would suggest to keep the naming `setPlotFormat` even if it's not optimal. Or: 
* export both functions (`setPlotFormat `and `setDefaultPlotFormat`)
* `setPlotFormat `should just call `setDefaultPlotFormat`

Yuri05 (approved):

Yuri05:
looks good, just something wrong in the vignettes:
```
--- re-building 'pk-parameters.Rmd' using rmarkdown
Quitting from lines 91-123 (pk-parameters.Rmd) 
Error: processing vignette 'pk-parameters.Rmd' failed with diagnostics:
tools::buildVignettes: argument 'object' is of type 'character', but expected 'numeric, or integer'!
--- failed re-building 'pk-parameters.Rmd'
```

pchelle:
> 
> 
> looks good, just something wrong in the vignettes:
> 
> ```
> --- re-building 'pk-parameters.Rmd' using rmarkdown
> Quitting from lines 91-123 (pk-parameters.Rmd) 
> Error: processing vignette 'pk-parameters.Rmd' failed with diagnostics:
> tools::buildVignettes: argument 'object' is of type 'character', but expected 'numeric, or integer'!
> --- failed re-building 'pk-parameters.Rmd'
> ```

On my computer, there is no error when building the vignette. That's why I have issues in troubleshooting what's wrong.
The error message definitely comes from `validateIsNumeric`, `validateIsInteger`, or `validateIsPositive` after the last commit I was expecting that it would give the name of the object instead of 'object' but it did not work...
From the tests I tried, it may come from `formatNumerics` or from the functions called in the workflow initialization of default settings - e.g. validateIsInteger(numberOfCores).

pchelle:
I have identified where the issue happens, but I don't know why it happens only here and now.
The vignette fails during the unit conversion of the pk parameters: `toUnit`/`toBaseUnit` expect 'numeric' values but got 'character' instead. 
After a quick check, there are some NaN values in the exported data and I am wondering if `readr::read_csv` now understands those as characters since appveyor uses v2.0.0 of `readr` while my computer uses v1.3.1.

Yuri05:
@pchelle Cool that you found it - well done! PR ready to merge then?
How were those NaN values stored in the CSV: as "NaN" string?

Did not find a hint in readr release notes why it changed (https://cran.r-project.org/web/packages/readr/news/news.html)
In case we have some trouble with the new version: readr-developers kept the possibility to force the usage of the previous major version of readr (version one) - s. the first section in the link above.

@msevestre @PavelBal FYI

pchelle:
> 
> 
> @pchelle Cool that you found it - well done! PR ready to merge then?
> How were those NaN values stored in the CSV: as "NaN" string?
> 
> Did not find a hint in readr release notes why it changed (https://cran.r-project.org/web/packages/readr/news/news.html)
> In case we have some trouble with the new version: readr-developers kept the possibility to force the usage of the previous major version of readr (version one) - s. the first section in the link above.
> 
> @msevestre @PavelBal FYI

Yes, finally ready to merge. I was not able to understand what was wrong with the build fail of centos7 v4.1.0 (issue in the tests but it seems that ospsuite was not able to run a model...)

In the PKAnalysis results file, they were indeed stored as NaN. There was also an NA value for the unit (which was a fraction). 
I can try some tests of readr v2.0.0 if you want me to investigate more.

Yuri05:
>I was not able to understand what was wrong with the build fail of centos7 v4.1.0 (issue in the tests but it seems that ospsuite was not able to run a model...)

Seems to be the same issue as in https://github.com/Open-Systems-Pharmacology/OSPSuite-R/issues/549
Happens from time to time, and is probably either rClr- or Mono-Bug under Linux.

>In the PKAnalysis results file, they were indeed stored as NaN. There was also an NA value for the unit (which was a fraction).

Can you attach this PKAnalysis results file? While it could happen that we have NaN as value, NA for unit is surprising.

pchelle:
> 
> 
> > I was not able to understand what was wrong with the build fail of centos7 v4.1.0 (issue in the tests but it seems that ospsuite was not able to run a model...)
> 
> Seems to be the same issue as in [Open-Systems-Pharmacology/OSPSuite-R#549](https://github.com/Open-Systems-Pharmacology/OSPSuite-R/issues/549)
> Happens from time to time, and is probably either rClr- or Mono-Bug under Linux.
> 
> > In the PKAnalysis results file, they were indeed stored as NaN. There was also an NA value for the unit (which was a fraction).
> 
> Can you attach this PKAnalysis results file? While it could happen that we have NaN as value, NA for unit is surprising.

[A-PKAnalysisResults.zip](https://github.com/Open-Systems-Pharmacology/OSPSuite.ReportingEngine/files/6923719/A-PKAnalysisResults.zip)
When I open it using Excel, pk parameter values are plain NaNs while unit is an empty cell. NaN and NA were obtained after I read it using `readr::read_csv`

Yuri05:
>When I open it using Excel, pk parameter values are plain NaNs while unit is an empty cell. NaN and NA were obtained after I read it using readr::read_csv

Ok. I was afraid that NA unit would cause errors - but it's the same with the readr version 1 - and it seems to work (if I define a fraction PK parameter with the empty unit "" - i get correct results.

Just to be sure: can you execute the workflow attached with readr 2.0 and check if PK analysis and sensitivity values are correctly produced in the report?
[test_readr_NA.zip](https://github.com/Open-Systems-Pharmacology/OSPSuite.ReportingEngine/files/6937724/test_readr_NA.zip)
This is how the report looks like with readr 1.3.1:
[Report.docx](https://github.com/Open-Systems-Pharmacology/OSPSuite.ReportingEngine/files/6937735/Report.docx)

PR #538, opened 2021-09-14
Fixes #537 update packates

### LINKED ISSUE #537: Update to latest version of tlf and ospsuite-r

msevestre:


msevestre:
FYI @pchelle @abdullahhamadeh This PR will have the latest code regarding also the concurrent mode (that I showed this morning)

Yuri05:
@msevestre Can you also update the ospsuite-version for **linux** builds?
https://github.com/Open-Systems-Pharmacology/OSPSuite.ReportingEngine/tree/develop/.github/workflows/install_packages_bionic.sh#L4 and
https://github.com/Open-Systems-Pharmacology/OSPSuite.ReportingEngine/tree/develop/.github/workflows/install_packages_centos7.sh#L4 
(tlf update is not required because it's installed from github/develop)

msevestre:
@Yuri05 
Why don't we install ospsuite that way?
And why does it  work to use github here where it did not work for the main build??

Yuri05:
>Why don't we install ospsuite that way? 

Because it's much more complicated. tlf is pure R-code, ospsuite include different versions of C++ libs, which are stored in the referenced nuget packages. Probably doable, but it was much easier just to download the compiled package version.

>And why does it work to use github here where it did not work for the main build?? 

Good question. I don't know. Maybe because the build-docker-images are running on GitHub and not outside (on AppVeyor)?

msevestre:
Oh yes of course. Forgot that it was building everything!

Yuri05:
ok, merging for now. we can update the linux packages later

PR #543, opened 2021-09-15
Parallelize mean model simulations

abdullahhamadeh:


R/error-checks.R
```diff
@@ -85,6 +85,14 @@ validateIsPositive <- function(object, nullAllowed = FALSE) {
   }
 }
 
+validateIsNonNegative <- function(object, nullAllowed = FALSE) {
```
msevestre, line 88: Maybe call this ValidateIsStrictPositive?

R/error-checks.R
```diff
@@ -85,6 +85,14 @@ validateIsPositive <- function(object, nullAllowed = FALSE) {
   }
 }
 
+validateIsNonNegative <- function(object, nullAllowed = FALSE) {
+  validateIsOfType(object, c("numeric", "integer"), nullAllowed)
```
msevestre, line 89: use validateIsNumeric

R/parallel-simulation-task.R
```diff
@@ -0,0 +1,53 @@
...
+ParallelSimulationTask <- R6::R6Class(
+  "ParallelSimulationTask",
+  inherit = SimulationTask,
+  public = list(
+
+    #' @description
+    #' Create a `ParallelSimulationTask` object
+    #' @param ... parameters inherited from R6 class `SimulationTask` object
+    #' @return A new `SimulationTask` object
+    initialize = function(...) {
```
msevestre, line 12: hum...why is there  aconstructo that just call base?

R/utils.R
```diff
@@ -177,20 +177,31 @@ generateResultFileNames <- function(numberOfCores, folderName, fileName, separat
...
-  return(sim)
+
+  if (is.null(simulationSet$minimumSimulationEndTime)) {
+    return(simulation)
+  }
+
+  if (simulationSet$minimumSimulationEndTime > simulation$outputSchema$endTime) {
+    maximalIntervalIndex <- which(sapply(simulation$outputSchema$intervals, function(x) {
+      x$endTime$value
+    }) == simulation$outputSchema$endTime)[1]
```
msevestre, line 201: why [1] here? Is this obvious?

R/utilities-task.R
```diff
@@ -640,3 +644,37 @@ loadPlotPKRatioTask <- function(workflow, configurationPlan) {
...
+#' @return A `QualificationTask` object
+loadPlotDDIRatioTask <- function(workflow, configurationPlan) {
+  validateIsOfType(workflow, "QualificationWorkflow")
+  validateIsOfType(configurationPlan, "ConfigurationPlan")
+
+  # Time Profiles task is only active if the field is defined & not empty
+  active <- !isOfLength(configurationPlan$plots$DDIRatioPlots, 0)
+
+  taskFunction <- plotQualificationDDIs
+  nameFunction <- deparse(substitute(plotQualificationDDIs))
```
msevestre, line 663: `deparse(substitute` is everywhere... It's not like I don't like it, it's just that I have no idea what it does. Can we create a method that actually has a name that we understand, also in 6 months from now?

R/utilities-task.R
```diff
@@ -61,14 +61,18 @@ inactivateWorkflowTasks <- function(workflow, tasks = workflow$getAllTasks()) {
 loadSimulateTask <- function(workflow, active = TRUE, settings = NULL) {
   validateIsOfType(workflow, "Workflow")
   validateIsLogical(active)
-  taskFunction <- simulateModel
-  nameFunction <- deparse(substitute(simulateModel))
+
+  taskFunction <- simulateModelParallel
+  nameFunction <- deparse(substitute(simulateModelParallel)) # deparse(substitute(simulateModel))
```
msevestre, line 66: remove comment

R/utilities-task.R
```diff
@@ -61,14 +61,18 @@ inactivateWorkflowTasks <- function(workflow, tasks = workflow$getAllTasks()) {
...
   validateIsOfType(workflow, "Workflow")
   validateIsLogical(active)
-  taskFunction <- simulateModel
-  nameFunction <- deparse(substitute(simulateModel))
+
+  taskFunction <- simulateModelParallel
+  nameFunction <- deparse(substitute(simulateModelParallel)) # deparse(substitute(simulateModel))
+  simulationTaskInitializer <- ParallelSimulationTask$new
+
   if (isOfType(workflow, "PopulationWorkflow")) {
```
msevestre, line 69: Where do you check that there is not a mix of Population and Individual Simulations?
Also what about putting everything in one task? Why is the new task required

R/parallel-simulation-task.R
```diff
@@ -0,0 +1,53 @@
...
+      super$initialize(...)
+    },
+
+    #' @description
+    #' Run task and save its output
+    #' @param structureSets list of `SimulationStructure` R6 class
+    runTask = function(structureSets) {
```
msevestre, line 19: I don't see where stuff is actually being run in //. Where is runSimulations called?

msevestre (changes requested):

msevestre:
@abdullahhamadeh I am a bit lost with the changes. I don't see where stuff is actually called to run in parallel 
Maybe in next PR you could outline some of the logic of the code changed by commenting on the PR. It's very hard to review because you run the styler at the same time and there is a lot of noise

msevestre:
@abdullahhamadeh Also please rebase or merge your changes. This PR has over 43 commits....most of them have nothing to do with the actual code being fixed/changed

R/error-checks.R
abdullahhamadeh: I think this function is redundant now, because of a later commit that removed the code where it was called.  We can remove it.  But 0 values are valid in this function, which is why 'non-negative' is suggested here (or maybe 'non strict positive'?)

R/parallel-simulation-task.R
abdullahhamadeh: Just for the documentation, line 9: 'Create a `ParallelSimulationTask` object' instead of 'Create a `SimulationTask` object'.  Remove?

R/utilities-task.R
abdullahhamadeh: I am assuming that the simulation sets are either for individuals (mean model) or for populations.  

It is possible to implement some logic in the `runTask` function of the `SimulationTask` object and remove the `ParallelSimulationTask`, but I felt this way would be neater (even if it means an extra class).  

The `runTask` function receives a list of `structureSet` objects.  This list doesn't tell us immediately whether `structureSets` comes from a mean workflow or a population workflow.  We would need to inspect the `structureSet` elements individually to see if any one of them stores a path to a population file, and if so, simulate each `structureSet` simulation in sequence using the `simulateModelForPopulation` as the `taskFunction` (otherwise simulate all the mean model simulations in parallel using the `simulateModelParallel` task function).

Instead, here, the appropriate `taskFunction` is selected based on the workflow.  If it is a mean model workflow, use a `ParallelSimulationTask` which passes the `structureSets` list to `simulateModelParallel` (the `taskFunction`) for parallel simulation.  If it is a population workflow, then loop thru the `structureSets` list and run each simulation separately using the `simulateModelForPopulation` as the `taskFunction`.

R/parallel-simulation-task.R
abdullahhamadeh: The parallel simulation takes place in the `taskFunction` for this task, which is assigned in the function `loadSimulateTask`.  If a mean model workflow is to be simulated (in parallel) then the task function is `simulateModelParallel` (called from the `runTask` function of a `ParallelSimulationTask` object.  Otherwise the population simulation sets are run sequentially thru the `simulateModelForPopulation` function, which is called in a for loop (over all simulation sets) in the `runTask` function of a `SimulationTask` object.

R/utilities-task.R
abdullahhamadeh: merged these into one function, `getObjectNameAsString`

R/utilities-task.R
abdullahhamadeh: merged these into one function, getObjectNameAsString

R/error-checks.R
msevestre: oh yes. The other one should be StrictPositive lol. At any rate, if we don't need it...let's forget it

R/parallel-simulation-task.R
msevestre: I would... esp[ecially because `  #' @return A new `SimulationTask` object` is wrong :)
Constructor is a  standard concept when working with class. Let's remove unless it adds value

R/parallel-simulation-task.R
msevestre: ok thanks

R/utilities-task.R
abdullahhamadeh: done

R/utilities-task.R
msevestre: >I am assuming that the simulation sets are either for individuals (mean model) or for populations.

Is this the correct assumptions? Can't we have a workflow where things are mixed? Asking for a friend

>but I felt this way would be neater (even if it means an extra class).
I don't mind the extra class. Let's keep it

also you can use `if (isOfType(workflow, PopulationWorkflow))`  (e.g. class instead of string)

The only problem I can see if that if we add more Workflows, we need to update this code. So we lose some flexibility. But good for now

How many workflows do we have?

R/utilities-task.R
```diff
@@ -136,7 +140,7 @@ loadCalculateSensitivityTask <- function(workflow, active = FALSE, settings = NU
 
   return(SensitivityAnalysisTask$new(
     getTaskResults = runSensitivity,
-    nameTaskResults = deparse(substitute(runSensitivity)),
+    nameTaskResults = getObjectNameAsString(runSensitivity),
```
msevestre, line 143: reads better. You could even simply by getName, extractName etc.. But at least we know what's going on. great

R/utils.R
abdullahhamadeh: I think it depends on whether there could be multiple intervals in the outputSchema that have the same end time.  If not, then the [1] can be removed.

msevestre:
Looks good to me....a few merge conflicts to solve now.Once you merge the changes from develop, we should be good to go

Did you benchmark those changes bTW?

R/utilities-task.R
abdullahhamadeh: Done.  

We have: `MeanModelWorkflow` `PopulationWorkflow` `QualificationWorkflow`

So far workflows aren't mixed, they are either individual (mean/qualification) or population.  I'm guessing this may indeed change for qualification (@Yuri05 ?)

msevestre:
the build is failing. Some vignettes issues. Can you fix  this? Then we are good to merge

PR #546, opened 2021-09-16
542 output

pchelle:


msevestre:
@pchelle there are a tone of changes here that have nothing to do with the branch name. I know it's a bit more work but can we next time create one PR per change unless they are related changes?

R/error-checks.R
```diff
@@ -285,46 +285,60 @@ logErrorMessage <- function(message, logFolderPath = getwd()) {
...
-  # Units
-  # If unit is defined as a value in nonmemUnit
-  timeMapping <- dictionary[, dictionaryParameters$ID] %in% dictionaryParameters$timeID
-  dvMapping <- dictionary[, dictionaryParameters$ID] %in% dictionaryParameters$dvID
-
-  timeUnit <- as.character(dictionary[timeMapping, dictionaryParameters$nonmemUnit])
-  dvUnit <- as.character(dictionary[dvMapping, dictionaryParameters$nonmemUnit])
+  # Check of unit definitions:
+  # 1) unit defined in outptuts
+  dataUnit <- NULL
```
msevestre, line 322: isn't a list? E.g. `dataUnits`?

R/error-checks.R
```diff
@@ -486,3 +500,71 @@ validateHasUniqueValues <- function(data, dataName = "dataset", na.rm = TRUE, nu
...
 }
+
+validateIsIncludedInDataset <- function(columnNames, dataset, datasetName = NULL, nullAllowed = FALSE, logFolder = NULL) {
+  if (nullAllowed && is.null(columnNames)) {
+    return()
+  }
+  if (isIncluded(columnNames, names(dataset))) {
+    return()
+  }
+  if (is.null(logFolder)) {
```
msevestre, line 511: hum....should not the logErrorThenStop be responsifile for this (e.g. checkign the log folder) This is noise for the caller and something we will forget

R/error-checks.R
```diff
@@ -486,3 +500,71 @@ validateHasUniqueValues <- function(data, dataName = "dataset", na.rm = TRUE, nu
...
+  # If no unit defined at all
+  if (isOfLength(c(unit, unitColumn, dataUnit), 0)) {
+    stop(messages$errorNoDataUnit())
+  }
+  # If no unit defined by dictionray, all outputs need to define dataUnit
+  if (isOfLength(c(unit, unitColumn), 0)) {
+    if (!isSameLength(dataUnit, outputs)) {
+      stop(messages$errorNoDataUnitInOutputs())
+    }
+    return(invisible())
```
msevestre, line 558: sometime you return `return()` sometime `return(invisible())`
What's the difference?

R/error-checks.R
```diff
@@ -486,3 +500,71 @@ validateHasUniqueValues <- function(data, dataName = "dataset", na.rm = TRUE, nu
...
+
+  # Checks for errors
+  # If no unit defined at all
+  if (isOfLength(c(unit, unitColumn, dataUnit), 0)) {
+    stop(messages$errorNoDataUnit())
+  }
+  # If no unit defined by dictionray, all outputs need to define dataUnit
+  if (isOfLength(c(unit, unitColumn), 0)) {
+    if (!isSameLength(dataUnit, outputs)) {
+      stop(messages$errorNoDataUnitInOutputs())
```
msevestre, line 556: why stop and not logandstop? what's the difference?

R/messages.R
```diff
@@ -74,12 +81,43 @@ messages <- list(
     )
   },
 
-  errorUnitNotProvidedInMetaDataFile = function(observedMetaDataFile) {
+  errorNoDataUnit = function() {
```
msevestre, line 84: nice!

R/output.R
```diff
@@ -18,20 +18,21 @@ DataSelectionKeys <- list(
...
 #' @field dataSelection character or expression used to filter the observed data
+#' @field dataUnit Unit of the observed data
 #' @field dataDisplayName display name of the observed data
 #' @field pkParameters R6 class `PkParameterInfo` objects
 #' @field residualScale Scale for calculation of residuals as included in enum `ResidualScales`
 #' @export
 #' @import ospsuite
 Output <- R6::R6Class(
   "Output",
-  cloneable = FALSE,
```
msevestre, line 28: yep that the actual change for this PR :)

pchelle:
> 
> 
> @pchelle there are a tone of changes here that have nothing to do with the branch name. I know it's a bit more work but can we next time create one PR per change unless they are related changes?

I am so sorry. 
Because the `dataUnit` feature was also on the Output object I updated both together when I started my changes. `dataUnit` required the error checking of units to be improved so I ended up with all of these changes.
I can still split this into 2 PRs since the commit for clonable is separate. Just let me know.

R/error-checks.R
pchelle: From what I have understood, 3 different ways can be used to define the units of an observed dataset:
- Definition of units in dictionary file by filling the columns **datasetUnit** for **time** and **dv**
- Definition of units in dictionary by filling the column **datasetColumn** for **timeUnit** and **dvUnit** (this definition maps an observed data column for units and allow multiple units for a same dataset)
- Definition of **dv** units for the selected data of `Output` objects using the new field `dataUnit` (this definition maps subsets of observed data to their units and allow multiple units for a same dataset)

Since one definition is necessary, I need to throw an error in case of no definition (with a message reminding these 3 methods) and throw a warning in case of multiple definitions (reminding which methods get priority).

Since **datasetUnit** and **datasetColumn** are either a string value or `NA`, `NULL`, I wanted a similar format for the units defined by `Output` objects. So, `dataUnit` is first `NULL` , and if `Output` objects define `dataUnit`, I get an array of characters using `unlist(lapply(...))` (`sapply` can't be used here in case one of the `Output` objects did not define `dataUnit`, I'll get a list and not the character array).
Then, I can concatenate these definitions as an array of characters and check for its length.

R/error-checks.R
pchelle: I totally agree. This is one of the things I was thinking when I said the error handling should be improved.
Currently, `logErrorThenStop` has a default log folder of `getwd()` meaning that if used without a logFolder it can possible create a new log file somewhere out of the workflow folder.
And since some of the error checkings happen prior the workflow creation, we need to discuss if we want a log for them or directly print the error on the console.

R/error-checks.R
pchelle: The difference is the print of the return.
Adding `invisible` just enforce that `NULL` is not printed on the console whenever a `validate` method is used.

In more details, if you have these 2 functions: `f1 <- function(x){x}` and `f2 <- function(x){invisible(x)}`
Then, f1 prints what you return but not f2.
```
>f1(NULL)
NULL
>f1("a")
[1] "a"
>x <- f1("a") # when allocated, the output is not printed anyways
```
while
```
>f2(NULL)
>f2("a")
```

R/error-checks.R
pchelle: The error checking of unit is performed when creating a `SimulationSet` object prior the creation of a `Workflow` object.
At that point the user hasn't defined the folder where to save logs.
The question is do we save these errors in the current folder (possibly duplicating the log files) or do we just print the errors on the console ?

R/error-checks.R
msevestre: ok cool

R/error-checks.R
msevestre: ok . Can you create an issue for this (if it's not there already). I want the logging mechanism to be easy and transparent to use

R/error-checks.R
msevestre: ok so in the validate method, it should always be return(invisible()) correct? This is not the case I believe

R/error-checks.R
msevestre: I see. Sounds good

msevestre:
Merging. Thanks for the info. A few comments but not related to this PR directly

PR #559, opened 2021-09-28
558 digits

pchelle:


tests/data/mean-pk/A-pkAnalysis.csv
```diff
@@ -1,3 +1,3 @@
 "Path","Parameter","Value","Unit"
-"Organism|A|Concentration in container","C_max",2.85019207000732,"µmol/l"
-"Organism|A|Concentration in container","AUC_tEnd",3292.13500976563,"µmol*min/l"
+"Organism|A|Concentration in container","C_max",2.85,"µmol/l"
```
msevestre, line 2: Hum... Why are those values truncated? Number  of digits is only a display artefact not an `export` artefact. This looks suspicious to me

tests/data/pop-pk/Plasma-AUC_tEnd-ratio.csv
```diff
@@ -1,2 +1,2 @@
 "Population","N","5th percentile","25th percentile","50th percentile","75th percentile","95th percentile","mean","standard deviation","geo mean","geo standard deviation"
-"Pediatric",100,1.47152131715106,1.37784663543106,1.38997028155538,1.39477329670309,1.45921906303091,1.38807549126267,1.44344632433609,1.38818028405479,0.994688287352575
+"Pediatric",100,1.47,1.38,1.39,1.39,1.46,1.39,1.44,1.39,0.99
```
msevestre, line 2: same here. Table should be exported using full digits IMO
@Yuri05 thoughts?

msevestre (changes requested):
Also I can't seem to find where sprintf is being used in the diff

R/utilities-writing-report.R
```diff
@@ -1,393 +1,400 @@
...
+    check.names = FALSE,
+    #colClasses = "character",
+    fileEncoding = "UTF-8",
+    stringsAsFactors = FALSE
+  )
+  table <- formatNumerics(
+    table,
+    digits = digits %||% reEnv$formatNumericsDigits,
+    scientific = scientific %||% reEnv$formatNumericsScientific
+    )
```
pchelle, line 92: @msevestre `formatNumerics` is now applied only when added to the .md report.
The results exported in .csv have all the necessary digits.
Also, let me know if I should expose the `formatNumerics` method (which is now different from what is implemented in `ospsuite` package with no need for `nsmall`)

msevestre:
@pchelle ok going to merge this. I can't believe that we have to write such code to format digit properly in R to be honest
We have such a function in .NET and we should expose it via ospsuite-r instead. Good enough for now

PR #560, opened 2021-09-28
Update name of simulation file from configuration plan for qualification reports

abdullahhamadeh:


msevestre:
@abdullahhamadeh build is failing

tests/data/configuration-plan/simulation-mappings.csv
```diff
@@ -1,4 +1,4 @@
-project,simulation,path
-Midazolam,S1,Midazolam/S1/
-Midazolam,S2,Midazolam/S2/
-Itraconazole,S1,Itraconazole/S1/
+project,simulation,path,simulationFile
```
msevestre, line 1: @abdullahhamadeh What is this file here?  Is this creating on the fly based on the configuration file? and if yes...why?

abdullahhamadeh: @msevestre, it is read using `read.csv` to create a dataframe that will be compared to the `simulationMappings` to be read from [this minimal configuration file](https://raw.githubusercontent.com/Open-Systems-Pharmacology/OSPSuite.ReportingEngine/f1779e0283d4b1d3626761defa505e913aa420c9/tests/data/configuration-plan/minimal.json).  The dataframes should be identical.

msevestre: But why do we need to duplicate the info? I don't like having more than one source of truth

abdullahhamadeh: I see.  

At a minimum we want to check that the `ConfigurationPlan` class is creating a `simulationMappings` dataframe with the four required column names (`project,simulation,path,simulationFile`).

But ideally we would build the configuration file out of the csv.

msevestre: >But ideally we would build the configuration file out of the csv.

I think it should be build out of the json file not the csv. What tool is creating this csv?

abdullahhamadeh: Right now the CSV is created manually (right, @pchelle?).  Downside is that it would need to be adjusted if the json fields change (as in this PR).

The `simulationMappings` dataframe is created using the `simulationMappings` active method in the `ConfigurationPlan` class.

If we build out of the json, we are essentially recreating this active method (the one we want to test), no?

pchelle: The `SimulationMappings` is initially written as a list of lists in the `json` file.
The `ConfigurationPlan` flattens the info as a data.frame which is convenient when using methods such as `getSimulationPath` or `getSimulationResultsPath`.

The test was comparing for a reference `json` the flatten data.frame to a reference data.frame saved as a `csv` file

PR #575, opened 2021-10-28
WIP 540 plot ref obs data

pchelle:


TWendl:
define on simulation level is preferable rather than on workflow level

pchelle:
> define on simulation level is preferable rather than on workflow level

If so, it might be better to define a new optional field(s) for `PopulationSimulationSet` objects then

@TWendl , could you provide me with more information about the feature, especially:
* Is this definition by simulation only for inclusion of reference observed data ? 
* Is it also for inclusion of reference simulated data ? 
* Do we want/need to split between reference observed and simulated data ?

msevestre:
@TWendl 
@pchelle needs some feedback on the implementation, Can you get back to him when you have time?

Yuri05:
@pchelle After discussion with @TWendl : * Reference **simulation** data should always be plotted; we don't need extra options for that.

PR #582, opened 2021-11-04
580 gof plot

pchelle:


R/qualification-comparison-time-profile.R
```diff
@@ -87,8 +87,8 @@ addOutputToComparisonTimeProfile <- function(outputMapping, simulationDuration,
     project = outputMapping$Project,
     simulation = outputMapping$Simulation
   )
-  simulation <- loadSimulation(simulationFile)
-  simulationResults <- importResultsFromCSV(simulation, simulationResultsFile)
+  simulation <- ospsuite::loadSimulation(simulationFile)
```
Yuri05, line 90: @pchelle This would be a candidate for `loadFromCache=true`, right?

R/qualification-gof.R
```diff
@@ -1,453 +1,298 @@
...
+  # Get simulation output
+  simulationFile <- configurationPlan$getSimulationPath(
+    project = outputMapping$Project,
+    simulation = outputMapping$Simulation
+  )
+  simulationResultsFile <- configurationPlan$getSimulationResultsPath(
+    project = outputMapping$Project,
+    simulation = outputMapping$Simulation
+  )
+  simulation <- ospsuite::loadSimulation(simulationFile)
```
Yuri05, line 108: here as well

R/qualification-comparison-time-profile.R
pchelle: I have added issue #583 for that because most (if not all) of the tasks are currently simply loading the simulation without any `loadFromCache=true`

PR #586, opened 2021-11-09
578 pk parameter offset

pchelle:


R/pk-parameter-info.R
```diff
@@ -17,10 +19,12 @@ PkParameterInfo <- R6::R6Class(
...
     #' @param pkParameter PK Parameter name for the simulation (e.g. `AUC_inf`)
     #' Tips: use ospsuite::allPKParameterNames() to get a list of available PK parameters
+    #' @param group Grouping identifier. Default group associate `pkParameter`.
     #' @param displayName display name for `pkParameter`
     #' @param displayUnit display unit for `pkParameter`
     #' @return A new `PkParameterInfo` object
     initialize = function(pkParameter,
+                          group = NULL,
                           displayName = NULL,
                           displayUnit = NULL) {
```
Yuri05, line 29: let's define "group" as the last argument, so that the old code does not break in case of unnamed arguments in calls of `PKParameterInfo$new()`

Yuri05:
Merging (created new issue for the typos: https://github.com/Open-Systems-Pharmacology/OSPSuite.ReportingEngine/issues/589)

PR #592, opened 2021-11-11
Fix failing build (install missing READR package) and disable progress when installing pandoc

Yuri05:


msevestre:
Where does this PAT comes from? 
I have fixed the build by downloading specific version of our components for now
Vignettes build is failing /cc @pchelle  This is why the build is red now

appveyor.yml
```diff
@@ -37,7 +37,7 @@ build_script:
   - Rscript -e "install.packages('https://github.com/Open-Systems-Pharmacology/OSPSuite-R/releases/download/v10.0.72/ospsuite_10.0.72.zip', repos = NULL, type = 'binary')"
   - Rscript -e "install.packages('https://ci.appveyor.com/api/buildjobs/lq7s2r98966wqpjh/artifacts/tlf_1.2.33.zip', repos = NULL, type = 'binary')"
```
msevestre, line 38: @Yuri05 I am installing the packages as zip here. THat;s why it works?

Yuri05:
>Where does this PAT comes from?
I have fixed the build by downloading specific version of our components for now
Vignettes build is failing /cc @pchelle This is why the build is red now

The PAT is my GH token encrypted via AppVeyour. I think any valid GH token would do here.
Vignettes were failing due to the missing package readr. Not sure why it was not installed with other dependencies (missing entry in DESCRIPTION?)

With that, the build process runs through.

appveyor.yml
Yuri05: Not sure I understand the question...

msevestre:
yes but the PAT is not required here because I am installing the package directly
I had tried this (my own PAT) but it did not work either. So if you want to be sure that it works, you need to revert my changes

appveyor.yml
msevestre: THE PAT File is only used when installing from github which I have removed a couple of hours ago
So the PAT is defacto not used here (as you can see, tlf and ospsuite-r are installed from packages and not from github. So the PAT changes are useless
To test if this works, you need to undo my changes from previous commit in develop
Also personal PAT expires after 30 days? This is not sustainable

msevestre (changes requested):

Yuri05:
>THE PAT File is only used when installing from github which I have removed a couple of hours ago
So the PAT is defacto not used here (as you can see, tlf and ospsuite-r are installed from packages and not from github. So the PAT changes are useless

OK, I see. I missed that commit (no PR?)
Will remove the PAT then.

Yuri05:
@msevestre ok, PAT removed and still works - very nice :)

pchelle:
> 
> 
> @msevestre @pchelle
> 
> > Where does this PAT comes from?
> > I have fixed the build by downloading specific version of our components for now
> > Vignettes build is failing /cc @pchelle This is why the build is red now
> 
> The PAT is my GH token encrypted via AppVeyour. I think any valid GH token would do here.
> 
> Vignettes were failing due to the missing package readr. Not sure why it was not installed with other dependencies (missing entry in DESCRIPTION?). So I installed it manually. With that, the build process runs through.

Did we know which function was using `readr` package ? (If I recall, only `ospsuite` was using `readr`)

pchelle (approved):
Just to make sure, does using the .zip also means we'll have to update the version of `tlf` and `ospsuite` every time they are updated ?

Yuri05:
OK, merging for now.

>Just to make sure, does using the .zip also means we'll have to update the version of tlf and ospsuite every time they are updated ?

In principle yes.
@msevestre we could also create timely unrestricted PAT and try with it. 
Or what we also could do: upload the .zip artifact as a copy without version number. E.g. not only **tlf_1.2.3.zip** but additionally **tlf.zip** (which is just a renamed version of **tlf_1.2.3.zip**). And then use the [permalink to the latest successfully build artifact](https://www.appveyor.com/docs/packaging-artifacts/#permalink-to-the-last-successful-build-artifact) in the appveyor.yml

PR #631, opened 2021-11-26
Force log plots to get ticks

pchelle:


R/utilities-plots.R
```diff
@@ -40,13 +40,39 @@ displayDimension <- function(dimension){
 #' @param x Values from which the limits are calculated
 #' @return A list of units for goodness of fit results
 #' @keywords internal
-autoAxesLimits <- function(x) {
+autoAxesLimits <- function(x, scale = NULL) {
```
msevestre, line 43: Should the default for scale be one of our scale?
Also the scale is missing from the list of parameter in the description

pchelle: Yes, sorry about that, I'll use `tlf::Scaling$lin` instead and document.
I was trying to troubleshoot the error that I guessed from the message since `devtools::check()` was working on my laptop

PR #633, opened 2021-11-29
WIP for exported plot dimensions based on Abdullah's code

pchelle:
@abdullahhamadeh ,
I would like to spread the code to the other plots/tasks, but I may need some insight to prevent breaking everything for DDI plots


R/utilities-plots.R
```diff
@@ -77,4 +77,77 @@ autoAxesTicksFromLimits <- function(limits) {
...
+#' Initial estimates based on Abdullah's tests
+#' TODO: improve this directly from tlf
+#' @param legendPosition The name of the legend position as defined by `tlf` enum `LegendPositions`
+#' @return A list of scaling values for `width` and `height`
+#' @keywords internal
+getLegendScalingFactors <- function(legendPosition = tlf::LegendPositions$outsideTop){
+  # Legend on the left/right sides: increase width
+  if(isIncluded(legendPosition, c(tlf::LegendPositions$outsideRight, tlf::LegendPositions$outsideLeft))){
+    return(list(width = 1.6, height = 1.2))
+  }
```
abdullahhamadeh, line 141: @pchelle : because this is a scaling that deals specifically with the legend, I think it would be better to scale the width by  4/3 (=1.6/1.2) and keep the height as is (ie scale the height by 1).  This is because the legend only takes up space horizontally.  If we need the 1.2x scaling, it can be introduced as an additional scale factor applied to the whole image elsewhere, in another function.

R/utilities-plots.R
```diff
@@ -77,4 +77,77 @@ autoAxesTicksFromLimits <- function(limits) {
...
+  # Legend on the top/bottom sides: increase height
+  if(isIncluded(legendPosition, c(tlf::LegendPositions$outsideTopLeft,
+                                  tlf::LegendPositions$outsideTop,
+                                  tlf::LegendPositions$outsideTopRight,
+                                  tlf::LegendPositions$outsideBottomLeft,
+                                  tlf::LegendPositions$outsideBottom,
+                                  tlf::LegendPositions$outsideBottomRight))){
+    return(list(width = 1.2, height = 1.4))
+  }
```
abdullahhamadeh, line 150: and here, scale the height only by 1.4/1.2

R/utilities-plots.R
```diff
@@ -77,4 +77,77 @@ autoAxesTicksFromLimits <- function(limits) {
...
+  # Otherwise use these default values
+  return(list(width = 1.2, height = 1.2))
```
abdullahhamadeh, line 152: otherwise: `return(list(width = 1.0, height = 1.0))`

abdullahhamadeh (changes requested):
minor changes re legend scalings for now

R/utilities-plots.R
```diff
@@ -79,6 +79,79 @@ autoAxesTicksFromLimits <- function(limits) {
...
+#' that accounts for possible shrinking of the plot panel due to the addition of the legend
+#' @return A `PlotConfiguration` object
+#' @keywords internal
+getPlotConfigurationFromPlan <- function(plotProperties, plotType = NULL, legendPosition = reEnv$theme$background$legendPosition) {
+  # Define the appropriate configuration from plotType
+  # by creating expression: "tlf::<plotType>PlotCOnfiguration$new()"
+  plotConfiguration <- eval(parse(text = paste0("tlf::", plotType, "PlotConfiguration$new()")))
+  
+  # Set properties from FontAndSize
+  fonts <- plotProperties$FontAndSize$Fonts
```
abdullahhamadeh, line 97: @pchelle  In the version of Qualification-DDI-CYP3A4 I have there is no `FontAndSize` field under the plot settings.  Has this changed for all configuration plans?

image

pchelle: Good catch !
In the global settings there is not a `FontAndSize` field, but there is one in local settings for time profile plots.
I'm going to try to generalize that and account for both

pchelle: Actually, the Plot Settings here are global settings sent to the current theme properties when loading/updating the configuration plan (see [`ConfigurationPlan$updateTheme()`](https://github.com/Open-Systems-Pharmacology/OSPSuite.ReportingEngine/blob/a5492d694ad0a384561c022cf3633d8c364cfb06/R/configuration-plan.R#L251)).

The function `getPlotConfigurationFromPlan()` is however called locally. In such case, I only found plot settings for time profile plots which looked like this:
image

Since some settings - e.g. for axes, background or legend - were sometimes defined outside `FontAndSize`, I used the upper level field as input of `getPlotConfigurationFromPlan`. This would allow more flexibility if new fields or field names are used in different plots. 

I did not apply the scaling of font at the theme level, I am going to apply it anyways when re-allocating the font properties.

abdullahhamadeh: just tried it - yes this solution works great!

R/utilities-plots.R
abdullahhamadeh: At least for the DDI, I think we will need the additional 1.2x stretch for the plot width and height only.  I don't think it is needed for the font sizes.  So can we introduce, for example

`reEnv$plotScaleFactor <- 2.4`

and then 
```
plotConfiguration$export$width <- reEnv$plotScaleFactor*legendScaling$width*width
plotConfiguration$export$height <- reEnv$plotScaleFactor*legendScaling$height*height
```

PR #634, opened 2021-11-29
Fix ratio calculation due to display unit of obs and sim data in table

### LINKED ISSUE #637: PK ratio calculation does not use the appropriate unit for simulated data
Now that there is a display unit for obs and sim data, 
the ratio is messed up because still using the base unit of simulation

pchelle:
Fixes #637


Yuri05:
@pchelle Do we have an issue for this? (just not quite sure what this fix is about)

pchelle:
There was not yet, I just noticed that when I re-ran the pediatric test

pchelle:
> @pchelle Do we have an issue for this? (just not quite sure what this fix is about)

My bad, I went too fast on that PR.
I just referenced one and linked to the PR

PR #640, opened 2021-12-03
604 re theme

pchelle:


pchelle:
@msevestre and @Yuri05 , 
the error is because `readr` was not installed
```r
Rscript -e "install.packages('readr', repos = 'http://cran.us.r-project.org', type='win.binary')"
Installing package into 'C:/RLibrary/4.0'
(as 'lib' is unspecified)
Warning: unable to access index for repository http://cran.us.r-project.org/bin/windows/contrib/4.0:
  cannot open URL 'http://cran.us.r-project.org/bin/windows/contrib/4.0/PACKAGES'
Warning message:
package 'readr' is not available as a binary package for this version of R
A version of this package for your version of R might be available elsewhere,
see the ideas at
https://cran.r-project.org/doc/manuals/r-patched/R-admin.html#Installing-packages 
travis-tool.sh run_tests
```

NAMESPACE
```diff
@@ -121,7 +121,6 @@ export(setDefaultNumericFormat)
 export(setDefaultPlotFormat)
 export(setDefaultStairstep)
 export(setDefaultTheme)
-export(setDefaultThemeFromJson)
```
Yuri05, line 124: `setDefaultThemeFromJson` removed because the function is available in the **tlf**?

R/reportingengine-env.R
```diff
@@ -132,28 +132,25 @@ setDefaultNumericFormat <- function(digits = NULL, scientific = NULL) {
...
 #' @export
-setDefaultTheme <- function(theme) {
-  validateIsOfType(theme, "Theme")
-  reEnv$theme <- theme
+setDefaultTheme <- function(theme = NULL) {
+  validateIsOfType(theme, "Theme", nullAllowed = TRUE)
+  reThemeFile <- system.file("extdata", "re-theme.json", package = "ospsuite.reportingengine")
+  # If the RE default is not found, the template theme from tlf will be used instead
+  reEnv$theme <- ifEqual(reThemeFile, "", reEnv$theme, tlf::loadThemeFromJson(reThemeFile))
+  reEnv$theme <- theme %||% reEnv$theme
```
Yuri05, line 150: I would extract the lines 147-149 into a separate function (e.g. `setDefaultRETheme()`) and then replace the line 150 with
`reEnv$theme <- theme %||% setDefaultRETheme()`
Because now if a not null `theme `is passed: lines 147-149 are executed for nothing

R/reportingengine-env.R
```diff
@@ -132,28 +132,25 @@ setDefaultNumericFormat <- function(digits = NULL, scientific = NULL) {
...
+#' If `NULL`, the current theme is re-initialized to the reporting engine default
 #' @export
-setDefaultTheme <- function(theme) {
-  validateIsOfType(theme, "Theme")
-  reEnv$theme <- theme
+setDefaultTheme <- function(theme = NULL) {
+  validateIsOfType(theme, "Theme", nullAllowed = TRUE)
+  reThemeFile <- system.file("extdata", "re-theme.json", package = "ospsuite.reportingengine")
+  # If the RE default is not found, the template theme from tlf will be used instead
+  reEnv$theme <- ifEqual(reThemeFile, "", reEnv$theme, tlf::loadThemeFromJson(reThemeFile))
```
Yuri05, line 149: As far as I understand the line 149, it says:
```
if reThemeFile is empty: do nothing
otherwise: load theme from reThemeFile
```
It's not wrong, however it would be more clear for me if the code would be written in the form
if reThemeFile is NOT empty: load theme from reThemeFile

Yuri05:
@pchelle I triggered a rebuild of your PR on AppVeyour (no code change) and it worked. Seems to be  just a temporary problem (connection between appveyour and CRAN)

NAMESPACE
pchelle: I removed it because I indeed thought so, but actually it is not the case.
So, I'll put it back on RE

R/reportingengine-env.R
pchelle: I agree, I'll change the code accordingly

Yuri05 (approved):

Issue #524, opened 2021-08-29
Inherited classes and Intellisense


Yuri05:
At the moment when user creates an instance of **inherited** class: initializer parameters of the **base** class are not shown via intellisense, so that it's hard to create an object without looking into the base class documentation.

E.g. when I create a **MeanModelWorkflow**, I need to look into the `Workflow$new()` documentation:
![grafik](https://user-images.githubusercontent.com/25061876/131248880-61f3c7fa-93c1-47d7-b140-9751d91abe42.png)![grafik](https://user-images.githubusercontent.com/25061876/131248947-682180bc-7611-45f4-bc3c-f4cd245cceee.png)

### If not: should we **explicitely** list all base class parameters in the definition of the initializer of an inherited class?
E.g. at the moment this is done for the **PopulationWorkflow**-class, so that all initializer parameters are shown.
From the code perspective it's duplication, but from the user perspective it's much easier to use:
![grafik](https://user-images.githubusercontent.com/25061876/131248841-21cef259-1ceb-4003-b0b4-221a4bb9542c.png)
https://github.com/Open-Systems-Pharmacology/OSPSuite.ReportingEngine/blob/ea2f843e29451bd6e65183e2d970e2329b15ee3b/R/population-workflow.R#L28-L50

It will become especially critical in the TLF when defining a plot configuration for different plot types. The base class (`PlotConfiguration`) has tons of parameters in the initializer, which are all not shown when creating a configuration of different plot types.

![grafik](https://user-images.githubusercontent.com/25061876/131249077-1d887b56-3a2b-4aa1-ba33-3bad45d84f93.png)  ![grafik](https://user-images.githubusercontent.com/25061876/131249152-f3a2efff-b9df-4334-9937-87ffe6542a23.png)

msevestre:
@PavelBal Do you maybe also know of a way to do this ?

PavelBal:
No, unfortunately not. The only way I am aware of is to list all the necessary arguments in the inherited class.

abdullahhamadeh:
There may be a solution to this using the `formals()` function, but there is one hurdle to be resolved.

The idea is to build a closure that acts as the child class constructor.  This function first runs the parent initialize function and then an 'extended' `initialize` function that is specific to the child class.  The parent initialize function is defined outside the R6 class definition of the parent class.

For this example, we have a constructor for the parent class

```
parentInitializer <- function(par1,par2){
[...]

PR #644, opened 2021-12-08
WIP intellisense with inheritance for MeanModelWorkflow initializer

abdullahhamadeh:
addresses #524 

demo using "tests/dev/test-mean-nonparallel.R"

image


abdullahhamadeh:
@pchelle, as discussed, I have tried simplifying the changes in this PR by simply extending the formals of the `MeanModelWorkflow` initializer, using  
`formals(MeanModelWorkflow$public_methods$initialize) <-  c(formals(Workflow$public_methods$initialize),formals(MeanModelWorkflow$public_methods$initialize))`
following the definition of the `MeanModelWorkflow` class.

What I'm seeing from initial testing, is that this isn't enough to modify the behavior.   The formals of `MeanModelWorkflow$new` do change, but the initializer doesn't seem to pass the arguments as expected.  I ran

```
rm(list = ls())
devtools::load_all(".")
formals(MeanModelWorkflow$public_methods$initialize) <-  c(formals(Workflow$public_methods$initialize),
                                                           formals(MeanModelWorkflow$public_methods$initialize))
ms <- SimulationSet$new(
  simulationSetName = "meansim",
  simulationFile = "./tests/data/input-data/individualPksimSim.pkml",
  outputs = Output$new(path = "Organism|Heart|Interstitial|smarties|Concentration")
)
mwf <- MeanModelWorkflow$new(simulationSets = list(ms), workflowFolder = "./ex")
```
which gave this error:

image

msevestre:
Can you rebase/merge on develop so that we don't have so much noise from other PR

msevestre:
Hey @abdullahhamadeh 
What are we trying to do here? It looks like you are writing some kind of meta framework around the default initialization? As this been approved by @Yuri05 ? I just want to be sure we are not working on nice to have features now if there are still some prio 1 to implement

abdullahhamadeh:
Hi @msevestre 

#524 is labelled high priority - is this not the case?

I did discuss it with @Yuri05 at the jour fixe two weeks ago, and the idea is to extend this tool only to initializers that will be used by the typical user (ie no need to implement it to classes that are only used internally, by the developers). 
Right now we have intellisense for :
image

image

in addition to `MeanModelWorkflow$new`

The package installs and gives the intended behavior with these changes, but I am having trouble with documentation.  `devtools::document()` gives a bug:

image

which I have traced down to the use of the `formals` function.  In addition, documenting the parameters of the initializers that are to be created this way is not straightforward.  So this PR is something that I wanted to discuss at the next meeting.

msevestre:
Ok I was not at the meeting. I am wondering if all the efforts is worth it. If @Yuri05 approved of this approach, I am fine. My train of thoughts was more along the way of doing it by hand since it's supposed to be only for the public class and not for all classes.

abdullahhamadeh:
Yes I totally agree if we only restrict this to the few cases I mentioned.  I think the benefits will be more obvious if/when we extend this to public classes in `tlf` where we sometimes have more than one level of inheritance (see, eg, `DDIRatioPlotConfiguration`).   Then, any changes in arguments of a parent class constructor will automatically appear in the code completion of any child/grandchild class constructors.

msevestre:
Will we move this to ospsuite.utils then?

PR #650, opened 2021-12-18
WIP migration to ospsuite.utils

abdullahhamadeh:


Yuri05:
@abdullahhamadeh Any reason, why it is sometimes `ospsuite.utils::enum`, `ospsuite.utils::isNotNull` etc. and sometimes just `enum`, `isNotNull`, ... ?

abdullahhamadeh:
> @abdullahhamadeh Any reason, why it is sometimes `ospsuite.utils::enum`, `ospsuite.utils::isNotNull` etc. and sometimes just `enum`, `isNotNull`, ... ?

WIP :)

abdullahhamadeh:
@Yuri05 I think this is steady state now, so good to merge

R/output.R
```diff
@@ -61,19 +61,19 @@ Output <- R6::R6Class(
...
-      ifnotnull(displayName, validateIsOfLength(displayName, 1))
-      ifnotnull(displayUnit, validateIsOfLength(displayUnit, 1))
-      ifnotnull(dataUnit, validateIsOfLength(dataUnit, 1))
-      ifnotnull(dataDisplayName, validateIsOfLength(dataDisplayName, 1))
-      validateIsOfType(dataSelection, c("character", "expression"), nullAllowed = TRUE)
-      ifnotnull(dataSelection, validateIsOfLength(dataSelection, 1))
-      validateIsOfType(c(pkParameters), c("character", "PkParameterInfo"), nullAllowed = TRUE)
-      
+
+      ospsuite.utils::validateIsString(path)
```
msevestre, line 65: I have to say that I don't like the way it reads. So much noise.
Is there a way to not have to add ospsuite.utils:: all the time?
@abdullahhamadeh @IndrajeetPatil ? I am going to merge for now but I'd like to discuss this

IndrajeetPatil: No, this is not at all necessary. We can just have one `#’ import ospsuite.utils` tag in roxygen documentation somewhere, and we won't need to use namespace (`::`).

I didn't do so in my PR in `{ospsuite}` because these functions used to live there and I wanted to be absolutely clear from which namespace these functions were now coming from. 

I will remove the `ospsuite.utils::` namespace calls in PR to ospsuite package soon. The same can also be done in the current repo.

msevestre: Do you think this is good practice or not? It's more a question at that stage;-)

abdullahhamadeh: > Do you think this is good practice or not? It's more a question at that stage;-)

I was [following this](https://r-pkgs.org/namespace.html#import-r).   Ended up using @importFrom for `%||%` only.

abdullahhamadeh: Also, we use ospsuite:: a lot.  Would be good to be consistent whichever way we go...

msevestre: Good point..... 
I like reading ospsuite::loadSimulation.... because I know the logic is deferred to another package...hum

PR #653, opened 2022-01-10
Fixes #60 multiple observed data in comparison time profile and GOF plots

### LINKED ISSUE #60: Enable more than one observed data set in Qualification Plan-ComparisonTimeProfilePlots and GOF Plots
In the Qualification Plan for DDI, section `ComparisonTimeProfilePlots` I'd like to have the option to include more than just one observed data set, e.g. something like:

Annotation 2020-02-26 160007

pchelle:
Would it be possible to get a test with multiple observed data ?
I want to make sure it is `ObservedData` and not `ObservedDataSets`

### DIFF
--- a/R/qualification-comparison-time-profile.R
+++ b/R/qualification-comparison-time-profile.R
@@ -6,7 +6,7 @@
 #' @return list with `plots` and `tables`
 #' @import tlf
 #' @import ospsuite
-#' @importFrom ospsuite.utils %||%
+#' @import ospsuite.utils
 #' @keywords internal
 plotQualificationComparisonTimeProfile <- function(configurationPlan,
                                                    logFolder = getwd(),
@@ -18,7 +18,7 @@ plotQualificationComparisonTimeProfile <- function(configurationPlan,
 
     # Get axes properties (with scale, limits and display units)
     axesProperties <- getAxesProperties(timeProfilePlan$Axes) %||% settings$axes
-    if (ospsuite.utils::isOfLength(axesProperties, 0)) {
+    if (isOfLength(axesProperties, 0)) {
       # TODO Centralize messaging of configurtion plan errors and warnings
       logWorkflow(
         message = paste0(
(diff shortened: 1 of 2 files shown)


Yuri05:
@pchelle Test attached
One of the time profile comparison plots has observed data as an array (and yes, it is `ObservedData`, not `ObservedDataSet `:))
[qualification_plan.zip](https://github.com/Open-Systems-Pharmacology/OSPSuite.ReportingEngine/files/7840548/qualification_plan.zip)

Yuri05:
@pchelle It seems that the ospuite package requires additional dependency: **tidyr**
Can you add it to https://github.com/Open-Systems-Pharmacology/OSPSuite.ReportingEngine/blob/9776ad288a1d35ac106f6741ce0cd5c01d9faab8/.github/workflows/install_packages_cran.sh#L4 ?
Then Linux builds should pass

R/qualification-comparison-time-profile.R
```diff
@@ -136,90 +136,90 @@ addOutputToComparisonTimeProfile <- function(outputMapping, simulationDuration,
...
-      )
-
-      observedError$ymax <- observedValues + ospsuite::toUnit(
-        ospsuite::getDimensionForUnit(observedResults$metaData$error$unit),
-        observedResults$data[, 3],
-        targetUnit = axesProperties$y$unit,
-        sourceUnit = observedResults$metaData$error$unit,
-        molWeight = molWeight
+  # Loop on each observed dataset in OutputMappings
+  for (observedDataSet in outputMapping$ObservedData) {
```
Yuri05, line 140: @pchelle this would work with both string **value** and string **array** in the json configuration plan?

Yuri05:
@pchelle For the GOF plots it could be done in the same way as for comparison time profiles. Would you like to it in this PR or in another one?

R/qualification-comparison-time-profile.R
pchelle: Yes.
I have tested the example, it works with both.
It also worked when I tested the addition of a second observed data in the array

pchelle:
> @pchelle For the GOF plots it could be done in the same way as for comparison time profiles. Would you like to it in this PR or in another one?

I had a look at the GOF plot code, it is mirroring the comparison time profile code.
The fix will be exactly the same in the file `qualification-gof.R`, I will add it in this PR

Issue #16, opened 2019-11-27
Calculate Population Sensitivity Task


Yuri05:
# Inputs
* Output folder of the Simulate population task (with calculated simulation results)
* Output folder of the Calculate Population PK Parameters task (with calculated PK-Parameters)
* Output Folder for results (all results will be generated here)
* OPTIONAL Number of Cores
* List of populations (this also includes definition of PK-Parameters of interest)
* List of Model - Parameters of interest (DEfault: Empty)
  * If the input List of model parameters is empty: calculate sensitivities for all effectively used non-formula non-categorical parameters (s. https://github.com/Open-Systems-Pharmacology/OSPSuite-R/issues/220 for details)
* Settings (s. also #26)
  * Number of steps for sensitivity calculation (Default: 2)
  * Variation range  for sensitivity calculation (Default: 0.1)
  * Percentiles for individuals of interest (Default: `5, 50, 95`)
# Outputs
* For each combination of `{Population, PK-Parameter, Model output}`
  * 3 Individuals from the population are selected: corresponding to the 5th, 50th and 95th percentile of `{PK-Parameter, Model output}` (or to Xth, Yth, Zth percentile in case the default settings were chenaged)
  * For each of those 3 Individuals: sensitivity for all combinations {PK-Parameter, Model output} are calculated and stored in the output folder
* Report structure will be as following:
  * Population 1
    * Output 1
      * PK-Parameter 1
         * all sensitivity plots/tables for {Pop1, Output1, Pk-Parameter1}
      * PK-Parameter 2
        * ...
    * Output 2
    * ...
  * Population 2
  * ...
# Remarks
* Parallelization must be done via MPI (https://github.com/Open-Systems-Pharmacology/OSPSuite.ReportingEngine/issues/40)

PR #663, opened 2022-01-18
659 increase coverage

pchelle:


Yuri05:
On linux the error below comes. Any idea why?

```
#16 1224.4 ══ Failed ══════════════════════════════════════════════════════════════════════
#16 1224.4 ── 1. Failure (test-sensitivity.R:104:3): Population workflows generate appropri
#16 1224.4 list.files(refOutputFolder)[4] absent from list.files(file.path(workflowA$workflowFolder, "SensitivityResults"))
#16 1224.4 
```

pchelle:
> On linux the error below comes. Any idea why?
> 
> ```
> #16 1224.4 ══ Failed ══════════════════════════════════════════════════════════════════════
> #16 1224.4 ── 1. Failure (test-sensitivity.R:104:3): Population workflows generate appropri
> #16 1224.4 list.files(refOutputFolder)[4] absent from list.files(file.path(workflowA$workflowFolder, "SensitivityResults"))
> #16 1224.4 
> ```

This test checks if the files generated from the population sensitivity analysis are the same between the test and a reference (`expect_setequal`).
In the reference, there are 5 files: `<simulationSetName>-popSensitivityResultsIndex` (mapping for each PK parameter the patients in the population on which the analysis is performed) + 4 files named `<simulationSetName>-SensitivityAnalysisResultsIndividualId-<IndividualId>` (for IndividualIds 3, 10, 49 and 56)

Not exactly sure, but my guess would be that either not the same patients are identified for the population sensitivity analysis on Linux (which would render an error in this file names comparison) or that an unidentified file is generated in that folder.

Yuri05:
>Not exactly sure, but my guess would be that either not the same patients are identified for the population sensitivity analysis on Linux (which would render an error in this file names comparison) or that an unidentified file is generated in that folder.

ok, let's merge for now, but this is something we need to investigate. Will create a new issue

PR #671, opened 2022-01-20
Fixes #660 qualification workflow template into a function

### LINKED ISSUE #660: Qualification workflow template: convert to a function
It's much easier to automate the execution of multiple qualification workflows if it is triggered by a **function** and not by a **script**.
So let's transform the script _inst/extdata/qualification-workflow-template.R_ into a function with parameters:
* qualificationRunnerFolder (mandatory)
* pkSimPortableFolder (default: NULL)
* createWordReport (default: TRUE)

pchelle:
@Yuri05 let me know if qualificationPlanName should also be parametrized by the function


Yuri05 (changes requested):
let's keep the function createQualificationReport as before: as a template, which is stored under extdata, and not as an exported package function.
Otherwise people will start writing wrappers around the function to customize their workflows - and this is exactly what I would like to avoid.

pchelle:
> let's keep the function createQualificationReport as before: as a template, which is stored under extdata, and not as an exported package function. Otherwise people will start writing wrappers around the function to customize their workflows - and this is exactly what I would like to avoid.

Oh ! I think I get it now. 
If I understand, the idea is that users will load the template function and source their version of it if they need to run multiple qualifications.

Yuri05:
or if users need any workflow adjustments which are not part of the qualification plan

Issue #347, opened 2020-10-27
Estimate max. Cores per Session


Yuri05:
When running in "non-MPI" mode: estimate max. number of cores (threads) available in the user session and pass it to .NET
(use default if this estimation fails)

```
getAllowedCores <- function() {
  cores <- tryCatch({
    # get cpu allowance from files
    cfs_quota_us <- as.numeric(system("cat /sys/fs/cgroup/cpu/cpu.cfs_quota_us", intern = T))
    cfs_period_us <- as.numeric(system("cat /sys/fs/cgroup/cpu/cpu.cfs_period_us", intern = T))
    cfs_quota_us / cfs_period_us
  },
  error = function(cond) {
    message("Error getting allowed cpu cores")
    return(NULL)
  },
  warning = function(cond) {
    message("Warning getting allowed cpu cores")
    return(NULL)
  })
}
```

msevestre:
Return null??? What about all other use cases where this won't work?

Yuri05:
Depends how the caller will use it. If the caller e.g. checks for null and then decides what to do - that's fine.
If the caller passes the return value directly to the ospsuite-r as an option - it should return the default value (e.g. `parallel::detectCores() - 1` )

msevestre:
I dont think it should error or warn. This will only work with a very specific infrastructure and therefore should fail gracefully

Yuri05:
>I dont think it should error or warn. This will only work with a very specific infrastructure and therefore should fail gracefully

I agree

PR #677, opened 2022-01-27
Apply runSimulations of sequence of subsets of structureSets

abdullahhamadeh:


DESCRIPTION
```diff
@@ -22,7 +22,8 @@ Imports:
     R6,
     ggplot2,
     jsonlite,
-    ospsuite.utils
+    ospsuite.utils,
+    parallel
 Suggests:
```
abdullahhamadeh, line 27: any objections to adding `parallel`?  It is in validR

R/utilities-simulation.R
```diff
@@ -134,37 +134,68 @@ simulateModelOnCore <- function(simulation,
...
 simulateModelParallel <- function(structureSets,
                                   settings = NULL,
                                   logFolder = getwd()) {
-  simulations <- lapply(structureSets, function(set) {
-    re.tStoreFileMetadata(access = "read", filePath = set$simulationSet$simulationFile)
-    simulation <- loadSimulationWithUpdatedPaths(set$simulationSet)
+
+  simulationResults <- list()
+  maxSimulationsPerSubset <- settings$maxSimulationsPerCore*((settings$allowedCores %||% parallel::detectCores()) - 1)  #To be set in settings argument
+
```
abdullahhamadeh, line 160: @Yuri05 since you touched on this in #347 :
- is `settings$allowedCores` redundant here?  ie does `parallel::detectCores()` work on linux as in windows?
- does `settings$allowedCores` return  `parallel::detectCores()` or does it return `parallel::detectCores() - 1`

DESCRIPTION
msevestre: Why do you need this? We are trying to parallelize stuff in .net (albeit not very successfully at the moment). If you do in in R, we will introduce potentially too much parallelism

abdullahhamadeh: I just need access to the number of cores we can use (pls see comment below) to calculate the maximal number of simulations per subset.  `getAllowedCores()` is only for linux and returns `NULL` on windows. If I have access to that without this package then this can go.

msevestre: Why do we have getAllowedCores that is different than the method we have in ospsuite-r?
Typically we have this method
https://github.com/Open-Systems-Pharmacology/OSPSuite-R/blob/develop/R/ospsuite-env.R#L73
getOSPSuiteSetting ["numberOfCores"] will return what you need. I am wondering if we should just not simply rely on it as default? (and yes this is using parallel)

R/utilities-simulation.R
msevestre: Why do we have settings$allowedCores ? Isn't is somehow redundant with the cores one? And how can it be undefined then?

abdullahhamadeh: OK `getOSPSuiteSetting` should get rid of the redundancy here.  But, indeed, what is the need for `getAllowedCores` (assuming `parallel::detectCores()` is fine in linux).

Yuri05: `getAllowedCores()` will return a value only if RE is executed on a **Kubernetes Cluster on Linux** (to assure that RE does not take more cores than are assigned to the user in the cluster session). In all other cases it will return NULL.

So to get the number of cores we must do something like this:
```
numberOfCores = getAllowedCores() 
if (numberOfCores == NULL)
{
    numberOfCores = getOSPSuiteSetting("numberOfCores")
}

```

msevestre: I think that should be encapsulated in a function. And maybe getAllowedCore should be renamed so that we know that it's only a specific case?

Yuri05: >I think that should be encapsulated in a function.

agree

>And maybe getAllowedCore should be renamed so that we know that it's only a specific case?

Yes. But don't ask me for the new function name 😄

R/utilities-simulation.R
```diff
@@ -173,7 +173,7 @@ simulateModelParallel <- function(structureSets,
 
     simulations <- lapply(structureSetsSubset, function(set) {
       re.tStoreFileMetadata(access = "read", filePath = set$simulationSet$simulationFile)
-      simulation <- loadSimulationWithUpdatedPaths(set$simulationSet,loadFromCache = TRUE)
+      simulation <- loadSimulationWithUpdatedPaths(set$simulationSet,loadFromCache = FALSE)
       logWorkflow(
```
abdullahhamadeh, line 177: This is a temporary fix to the failing build until [this issue](https://github.com/Open-Systems-Pharmacology/OSPSuite-R/issues/733) is addressed.

msevestre: I don't understand. Are you running the same simulation twice?

msevestre: I can only reproduce if you are loading the SAME Simulation twice and running it in parallel. While it should not crash, I do not see any reason to do this as you are basically wasting CPU time.  
So the question is: Why do you need to run the same simulation using `runSimulations` multiple times?
Maybe @pchelle can answer this as well

abdullahhamadeh: As mentioned [here](https://github.com/Open-Systems-Pharmacology/OSPSuite-R/issues/733), this threw an error only when I changed `loadFromCache` to `TRUE`.  When it was `FALSE` previously, this vignette example worked fine.  We can discuss whether simulations should be loaded from cache at this point.

R/simulation-settings.R
```diff
@@ -43,6 +46,19 @@ SimulationSettings <- R6::R6Class(
       }
     },
 
+    #' @field maxSimulationsPerCore  Scale factor used in a parallel simulation.  The product of this scale factor and the number of allowable cores (allowedCores) sets the maximum number of simulations that may be run on one core.
+    maxSimulationsPerCore = function(value) {
+      if (missing(value)) {
+        private$.maxSimulationsPerCore
```
msevestre, line 52: return here so that you do not need the else

R/simulation-settings.R
```diff
@@ -43,6 +46,19 @@ SimulationSettings <- R6::R6Class(
       }
     },
 
+    #' @field maxSimulationsPerCore  Scale factor used in a parallel simulation.  The product of this scale factor and the number of allowable cores (allowedCores) sets the maximum number of simulations that may be run on one core.
+    maxSimulationsPerCore = function(value) {
+      if (missing(value)) {
+        private$.maxSimulationsPerCore
+      } else {
+        if (!is.null(value)) {
```
msevestre, line 54: I would not check for null. It's an error to set a value explicitelu to Null no?

R/utilities-simulation.R
```diff
@@ -134,6 +134,14 @@ simulateModelOnCore <- function(simulation,
   return(simulationResult)
 }
 
+
+#' @title defaultMaxSimulationsPerCore
+#' @description Default value for a scale factor used in a parallel simulation.  The product of this scale factor and the number of allowable cores (allowedCores) sets the maximum number of simulations that may be run on one core.
+#' @export
+defaultMaxSimulationsPerCore <- 2
```
msevestre, line 141: Don't we have an environement where we can put those type of global variables?

R/utils.R
```diff
@@ -139,11 +139,11 @@ generateResultFileNames <- function(numberOfCores, folderName, fileName, separat
 #' @param loadFromCache logical allows load from Cache option
 #' @return simulation object with pathIDs updated from simulationSet
 #' @export
-loadSimulationWithUpdatedPaths <- function(simulationSet, loadFromCache = FALSE) {
+loadSimulationWithUpdatedPaths <- function(simulationSet, loadFromCache = FALSE, addToCache = FALSE) {
```
msevestre, line 142: default TRUE for addToCache no?

abdullahhamadeh:
> @msevestre this can be merged now as discussed

didn't see your other comments - will address those

R/utilities-simulation.R
abdullahhamadeh: moved to reEnv$defaultMaxSimulationsPerCore

R/simulation-settings.R
abdullahhamadeh: done

R/simulation-settings.R
abdullahhamadeh: done

msevestre (commented):
Looks good

msevestre:
There are a few conflicts...can you resolve?

Yuri05:
@abdullahhamadeh Can you please rebase this PR from **develop**? 
The error https://github.com/Open-Systems-Pharmacology/OSPSuite.ReportingEngine/issues/698 supposed to be fixed. so after reabase I will retry with DDI workflows.

PR #715, opened 2022-02-23
WIP test addToCache = FALSE in simulationSet creation to reduce memory usage

abdullahhamadeh:


Yuri05:
@abdullahhamadeh @pchelle 
Hmm, with this PR 1 test fails. I suspect that we have a bug in MassBalance-Task if simulations are added to cache. Needs to be investigated.

```
== Failed tests ================================================================
  -- Failure (test-mass-balance.R:42:3): Saved mass balance time profiles have correct values --
  readObservedDataFile(...) not equal to readObservedDataFile(refOutputMassBalance).
  Attributes: < Component "row.names": Numeric: lengths (1928, 2169) differ >
  Component "Time": Numeric: lengths (1928, 2169) differ
  Component "Amount": Numeric: lengths (1928, 2169) differ
  Component "NormalizedAmount": Numeric: lengths (1928, 2169) differ
  Component "Legend": Lengths (1928, 2169) differ (string compare on first 1928)
  Component "Legend": 482 string mismatches
  
  [ FAIL 1 | WARN 66 | SKIP 0 | PASS 355 ]
```

Yuri05:
I think what happens here: **plotMassBalance** loads simulations from cache via `loadSimulationWithUpdatedPaths`:
https://github.com/Open-Systems-Pharmacology/OSPSuite.ReportingEngine/blob/f43b19233b027cf2c35dd55b1c165e7770d5b447/R/utilities-mass-balance.R#L18

And in the function `loadSimulationWithUpdatedPaths` the simulation end time is increased if the simulation set requires it
https://github.com/Open-Systems-Pharmacology/OSPSuite.ReportingEngine/blob/f43b19233b027cf2c35dd55b1c165e7770d5b447/R/utils.R#L161-L166

However **plotMassBalance** also adds simulation to cache (default argument of `loadSimulationWithUpdatedPaths`):
https://github.com/Open-Systems-Pharmacology/OSPSuite.ReportingEngine/blob/f43b19233b027cf2c35dd55b1c165e7770d5b447/R/utils.R#L142

So from just looking into the code I don't see why we have the difference now.

PR #726, opened 2022-03-02
Fixes #721 log warnings due to missing input

### LINKED ISSUE #721: Qualification / Calculate PK parameters: missing inputs are not catched
2nd issue for the example from https://github.com/Open-Systems-Pharmacology/OSPSuite.ReportingEngine/issues/720 
One of the simulations did not produce any output, thus the corresponnding Calculate PK parameters task could not find its input.
But no error or warning was shown (also not in the log files).

pchelle:


pchelle:
Since the path was not specified, the warnings were actually logged in working directory (default value) instead of workflow folder.

pchelle:
Currently, when known task inputs are missing only a warning is thrown to the user
Let me know if there is interest in throwing an error instead

Issue #736, opened 2022-03-17
Warnings: 'recursive' will be ignored as 'to' is not a single existing directory


Yuri05:
Running the example from https://github.com/Open-Systems-Pharmacology/OSPSuite.ReportingEngine/issues/732 produces a lot of warnings like below.

Is this problematic? If not: can we supress the warnings?
```
37: In file.copy(from = file.path(inputContentSubFolder, fileName),  ... :
  'recursive' will be ignored as 'to' is not a single existing directory
38: In file.copy(from = file.path(inputContentSubFolder, fileName),  ... :
  'recursive' will be ignored as 'to' is not a single existing directory
```

PR #737, opened 2022-03-18
734 remove warnings

pchelle:


R/configuration-plan.R
```diff
@@ -160,8 +160,7 @@ ConfigurationPlan <- R6::R6Class(
           # With recursive=TRUE, overwrite images
           file.copy(
             from = file.path(inputContentSubFolder, fileName),
-            to = file.path(outputContentSubFolder, fileName),
-            recursive = TRUE
+            to = file.path(outputContentSubFolder, fileName)
```
Yuri05, line 163: If my input ContentFolder looks like this:
```
Content
  -- file1.md
  -- images/
      -- pic1.png
      -- images2/
          -- pic2.png
```

Will it be copied correctly into the output content folder with this change?

pchelle: Does not work... the message is not there anymore but `file.copy` does not copy the sub files/folders

Yuri05: Well, then improve pls

R/configuration-plan.R
```diff
@@ -160,8 +160,7 @@ ConfigurationPlan <- R6::R6Class(
           # With recursive=TRUE, overwrite images
           file.copy(
             from = file.path(inputContentSubFolder, fileName),
-            to = file.path(outputContentSubFolder, fileName),
-            recursive = TRUE
```
msevestre, line 164: this is suspicious here. Why don't we copy recursively? Do we only have one layer?

pchelle: I had two issues using `file.copy`
1- I wanted to use `recursive` directly but it creates a subfolder instead of copying only the content
2- If the argument `from` of `file.copy` was a folder, it was trying to copy it as file and it did not work properly (which was the initial warning of issue #736 )

To fix both, I now use `recursive` in `list.files` and `list.dirs` to fetch the list of **all** files and folders. 
This way, I can create all the directories recursively from the folder names first and then copy directly the files at appropriate locations

R/configuration-plan.R
```diff
@@ -141,29 +141,34 @@ ConfigurationPlan <- R6::R6Class(
...
+      # Note that recursive is not needed here since list of folders used it
+      for(contentSubFolder in contentSubFolders){
+        dir.create(file.path(self$workflowFolder, contentSubFolder), showWarnings = FALSE)
+      }
+      # Copy all available files of Content within workflow folder
+      # Note that recursive is not needed here since list of files used it
+      for(contentFile in contentFiles){
+        file.copy(
+          from = file.path(self$referenceFolder, "Content", contentFile),
+          to = file.path(self$workflowFolder, contentFile)
```
Yuri05, line 170: is `to = file.path(self$workflowFolder, contentFile)` correct or should it be
`to = file.path(self$workflowFolder, "Content", contentFile)` ?

R/configuration-plan.R
```diff
@@ -141,29 +141,34 @@ ConfigurationPlan <- R6::R6Class(
...
-            to = file.path(outputContentSubFolder, fileName),
-            recursive = TRUE
+    #' @description If available, copy all files and folders of `Content` directory into `<workflowFolder>`
+    copyContentFiles = function() {
+      # If no Content available, does not need to copy anything
+      if(!dir.exists(file.path(self$referenceFolder, "Content"))){
+        return(invisible())
+      }
+      # Get list of folders within content as well as all the subfolders using recursive option
+      contentSubFolders <- list.dirs(path = file.path(self$referenceFolder, "Content"), full.names = FALSE, recursive = TRUE)
```
Yuri05, line 156: `file.path(self$referenceFolder, "Content")` is used 4 times in the function.
You could assign it to a varibale first, e.g.
`srcContentFolder = file.path(self$referenceFolder, "Content")`

R/configuration-plan.R
Yuri05: ok, should be good then

Yuri05 (approved):

PR #738, opened 2022-03-18
732 error observed units

pchelle:


R/utilities-qualification.R
```diff
@@ -262,9 +262,12 @@ QualificationPlotTypes <- c("GOFMergedPlots", "ComparisonTimeProfilePlots", "DDI
 #' @keywords internal
 separateVariableFromUnit <- function(variableUnitString) {
   splitVariableUnitString <- strsplit(x = sub("[ []", replacement = "", x = variableUnitString), split = "[][]")[[1]]
+  if(isOfLength(splitVariableUnitString, 1)){
+    return(list(name = splitVariableUnitString, unit = ""))
+  }
   return(list(
     name = trimws(splitVariableUnitString[1]),
-    unit = ifelse(test = is.na(x = tail(splitVariableUnitString, 1)), yes = "", no = tail(splitVariableUnitString, 1))
+    unit = tail(splitVariableUnitString, 1)
```
Yuri05, line 270: Which name/unit would the function return for the column name **Value [raw] from lab [mg/ml]** ?

pchelle:
```r
> separateVariableFromUnit("Value [raw] from lab [mg/ml]")
$name
[1] "Value"

$unit
[1] "mg/ml"
```

Yuri05: well, that's obviously wrong. Name should be `Value [raw] from lab`.
In general: 
* if the column name ends with `[something]`, then `something` is a unit and everything before is the name
* if the column name does not end with `[something]`, then the unit is empty and the column name is the name

pchelle: Is this the correct behaviour or do we want `unit`=`"raw"` ?

Yuri05: @pchelle The function does not work as expected.
```
> separateVariableFromUnit("a[xy]b")
$name
[1] "a"

$unit
[1] "b"
```
Please add unit tests and fix the function
* "Value" => name="Value", unit = ""
* "Value [unit]" => name="Value", unit = "unit"
* "Value [unit]  " => name="Value", unit = "unit"
* "Value [raw] 1 [unit]" => name="Value [raw] 1", unit = "unit"
* "Value [raw] [unit] " => name="Value [raw]", unit = "unit"
* "Value [raw] 1" => name="Value [raw] 1", unit = ""
* "[Value] [unit]" => name="[Value]", unit = "unit"
* ...

pchelle: @msevestre 
Just wondering if I am reinventing the wheel here
Is there a similar function in `{ospsuite}` package ? i.e one that splits between variable names and their units ?

msevestre: there is nothing readily available for you to use 

I will implement it in ospsuite-r for you to consume

Here is the code for reference

https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/blob/develop/src/OSPSuite.Infrastructure.Import/Core/DataFormat/AbstractColumnsDataFormat.cs#L157

and 

https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/blob/develop/src/OSPSuite.Infrastructure.Import/Core/DataFormat/AbstractColumnsDataFormat.cs#L162

Yuri05: >Why would we allow a name like this 

The names come AS IS from the names of observed data colums in PK-Sim. And those names come from the data sources. So if the user has a column name in his excel/csv/... data source - it will be propagated to PK-Sim and then to RE.

Here just an example: one of observed data files exported for RE in CYP3A4-DDI workflow.
[Abduljalil 2009, 500 mg po MD.csv](https://github.com/Open-Systems-Pharmacology/OSPSuite.ReportingEngine/files/8316833/Abduljalil.2009.500.mg.po.MD.csv)
While this example is surely something which could/should be fixed in the PK-Sim project: the more meaningful name containing square brackets still can occur

msevestre: here
https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/issues/1558

How should the API from R look like?

msevestre: ok a list with name and unit. Sounds good

msevestre: There is a PR on ospsuite.core fixing the issue and successfully implementing all the tests above
@pchelle Once this is merge, I will update ospsuite-r with a new method `extractNameAndUnit(string)` that you will be able to use right away and remove all that code

msevestre: @pchelle see here
https://github.com/Open-Systems-Pharmacology/OSPSuite-R/pull/838#issuecomment-1076549222

Define the method directly in your utility class. This is very specific to RE and does not belong to ospsuite-r
It will however use method from ospsuite.core directly

PR #757, opened 2022-04-01
Fixes #743 Pretty caption accounts for font size and plot width

### LINKED ISSUE #743: Qualification reports: Time profile legend looks "bad"
E.g. 
grafik

Compare output of the Matlab-RE for the same plot:
https://raw.githubusercontent.com/Open-Systems-Pharmacology/OSP-Qualification-Reports/master/DDI_Qualification_CYP3A4/images/003_3_Concentration-Time_Profiles/025_3_12_Fluconazole_-_Alfentanil_DDI/001_plotComparisonTimeProfile.png

* Question 1: The legend uses not the whole width of the plot area. Why? Even with this big legend font size we would avoid line breaks if it would be the case.
* Question 2: Was the font size from the report configuration plan used? Again looking on the Matlab plot: legend size is significantly smaller than the axes captions size; in the R plot the difference between those 2 sizes seems to be smaller
* Question 3: is it actually possible to display the legend entries without that "background"?
grafik

pchelle:
Algorithm is now based on plot width, font size and position of the element on which to use line breaks

Also, note that tlf PR #231 fixed issues about legend font size (by centralizing setLegendFont in initialiePlot) and symbols background (legend.in plotconfiguration-legend.R)


Yuri05:
@pchelle Do you have an example of a time profile plot with the new legend?

pchelle:
> @pchelle Do you have an example of a time profile plot with the new legend?

Sure, here are a few examples
comparison_time_profile_Chattopadhyay_2018_1

comparison_time_profile_Lutz_2008_4

Yuri05:
thanks. Looks good!
And any example with "very long" captions? :)

Yuri05:
And the legend looks so much better now without that "background" :)

pchelle:
> And any example with "very long" captions? :)

Yes, and I already flagged a new issue for `{tlf}`: alignment of legends and labels should be modifiable. Right now they are all centered.

8_time_profile_plot_Vancomycin_Healy1987_1gBID

Yuri05:
>Yes, and I already flagged a new issue for {tlf}: alignment of legends and labels should be modifiable. Right now they are all centered.

Sounds good.

PR #763, opened 2022-04-04
Fixes #761 add space between tables and their captions for word reports

pchelle:

PR #769, opened 2022-04-07
Fixes #767 Deviation in pop time profile translated as range plot

### LINKED ISSUE #767: Why don't we have a Range plot when using mean + STD?
image

In PK-Sim. we have a range plot

/cc @Yuri05 @pchelle @abdullahhamadeh

pchelle:


pchelle:
@msevestre 
In the configuration plan, the color is defined centrally in `Analysis`>`Fields`>`Color`
So the ranges and lines that summarize the simulations (in `Analysis`>`Statistics`) all get the same color

pchelle:
Linux don't work because latest update on `{ospsuite}` require package `{patchwork}`

msevestre:
Patchwork doesn't work on Linux?

pchelle:
> Patchwork doesn't work on Linux?

The list of packages installed in linux depends on `.github/workflows/install_packages_cran.sh`
I have included it in the previous PR #763

Yuri05:
Please rebase from develop

PR #775, opened 2022-04-08
Fixes #758 prevent plot area shrinking when saving plot

### LINKED ISSUE #758: Avoid "shrinking" of the plot area if the legend is above/below the plot
At the moment, if the legend is placed **above/below** the plot: `Legend.Height + PlotArea.Height = const`, meaning that the plot area is shrinking with every new legend entry.

Required is however: `PlotArea.Height = const` (or `PlotArea.Width : PlotArea.Height = const` )

pchelle:


pchelle:
Comparison of what the same test plot before and after using `updatePlotDimensions` :
# Before
before

# After
after

pchelle:
Works also on legend on the side

# Before
before

# After
after

Yuri05:
Please rebase from develop

Yuri05:
> Comparison of what the same test plot before and after using `updatePlotDimensions` :

Nice!
@sfrechen FYI

PR #815, opened 2022-05-13
814 report path

pchelle:
@Yuri05 
I exported the function `copyReport` so it can also be used outside of workflows if users are interested in moving/copying other markdown reports (which can be useful, if one wants to add only the report to github).
The function checks for the file links (that use``) within the report and copy them along the markdown report in the new location.


R/utilities-writing-report.R
```diff
@@ -709,3 +709,53 @@ updateTableNumbers <- function(fileContent, pattern = "Table:", replacement = "T
...
+#' @export
+copyReport <- function(from, to, keep = FALSE) {
+  validateIsFileExtension(from, "md")
+  validateIsFileExtension(to, "md")
+  validateFileExists(from)
+  validateIsLogical(keep)
+  # If from is 
+  if(from == to){
+    return(invisible())
+  }
```
Yuri05, line 726: dangerous, same path could be written in the different ways (especially under windows).
Better e.g. something like this:
```
if(tolower(normalizePath(paste0(from, "/"))) == tolower(normalizePath(paste0(to, "/")))){
...
```

MAybe there is more elegant way to do it in R, don't know.

R/utilities-writing-report.R
```diff
@@ -709,3 +709,53 @@ updateTableNumbers <- function(fileContent, pattern = "Table:", replacement = "T
...
+  # Get directories of reports
+  fromFolder <- dirname(from)
+  toFolder <- dirname(to)
```
Yuri05, line 729: is it possible that `from != to` but `fromFolder = toFolder ` ?
E.g. cann the function be called with like this: `copyReport("C:/temp/report1", "C:/temp/report2")`?

IF yes: we need additional check here, if `fromFolder `and `toFolder  `are equal. And if yes: no need to copy the images subfolder - only the report itself (md) must be copied

R/utilities-writing-report.R
pchelle: That's true.
Unfortunately, I don't think there is more elegant than that

R/utilities-writing-report.R
pchelle: Good point, I'll return without copying if folders are same

Yuri05:
The code looks good for me.
However when trying to execute the attached workflow it still crashes.
```
  reportFolder <- file.path(workingDirectory, "report")
  reportName <- file.path(reportFolder, "report.md")
```
[03_814.zip](https://github.com/Open-Systems-Pharmacology/OSPSuite.ReportingEngine/files/8692837/03_814.zip)

```
Error in cat(x, file = file, sep = c(rep.int(sep, ncolumns - 1), "\n"),  : 
  cannot open the connection
In addition: There were 35 warnings (use warnings() to see them)
> traceback()
6: cat(x, file = file, sep = c(rep.int(sep, ncolumns - 1), "\n"), 
       append = append)
5: write("", file = fileObject, sep = "\n")
4: resetReport(outputFile, logFolder)
3: mergeMarkdownFiles(mdFiles$appendices, file.path(self$workflowFolder, 
       self$reportFileName), logFolder = self$workflowFolder)
2: workflow$runWorkflow() at workflow.R#156
1: createQualificationReport
```

* please fix and add tests
* let's change the workflow template so that per default report is created in the report subfolder (like in the attached example)

Yuri05 (changes requested):
s. comment

pchelle:
> The code looks good for me. However when trying to execute the attached workflow it still crashes.
> 
> ```
>   reportFolder <- file.path(workingDirectory, "report")
>   reportName <- file.path(reportFolder, "report.md")
> ```

Template and workflow objects were updated.
I hope the behavior will be clearer using path instead of name:
- If using `reportFilePath` as in template now, the code will set/create the appropriate folder and change the name of the report (by splitting internally the path using `dirname()` and `basename()`) 

```r
worflow$reportFilePath <- "Report/new-report.md"

worflow$reportFilePath
> [1] "Report/new-report.md"

worflow$reportFolder
> [1] "Report"

worflow$reportFileName
> [1] "new-report.md"
```

- If using `reportFolder`, the code will set/create the appropriate folder and keep the report file name as is:
```r
worflow$reportFolder <- "Report"

worflow$reportFilePath
> [1] "Report/report.md"
```

- If using `reportFileName`, the code will only change the name of the file but not its folder (path separators should not be used for file name)
```r
worflow$reportFileName <- "new-report.md"

worflow$reportFilePath
> [1] "./new-report.md"
```

Yuri05:
[03_814.zip](https://github.com/Open-Systems-Pharmacology/OSPSuite.ReportingEngine/files/8699690/03_814.zip)

* Word report is not copied
grafik

* let's change the workflow template so that per default **report** is created in the report subfolder (like in the attached example)

* please add tests (with and without creating word report):
  * report path default
  * report path: same folder but other report file name
  * report path: other folder but same report name
  * report path: other folder and other report name

inst/extdata/qualification-workflow-template.R
```diff
@@ -77,8 +77,9 @@ createQualificationReport <- function(qualificationRunnerFolder,
   #' Default is no watermark. `Label` objects from `tlf` package can be used to specifiy watermark font.
   watermark <- ""
 
-  #' If not set, report created will be named `report.md` and located in  `reOutputFolder`
-  reportName <- file.path(reOutputFolder, "report.md")
+  #' If not set, report created will be named `report.md` and located in the worflow folder namely `reOutputFolder`
+  reportFolder = file.path(workingDirectory, "report")
```
Yuri05, line 81: let's use `<-` instead of `=`

inst/extdata/qualification-workflow-template.R
```diff
@@ -77,8 +77,9 @@ createQualificationReport <- function(qualificationRunnerFolder,
...
+  reportPath <- file.path(reportFolder, "My_report.md")
```
Yuri05, line 82: let's name the report per default `"report.md"` , not `"My_report.md"`

R/mean-model-workflow.R
```diff
@@ -100,14 +102,23 @@ MeanModelWorkflow <- R6::R6Class(
...
           wordConversionTemplate = self$wordConversionTemplate
           )
+        # Move report if a non-default path is provided
+        copyReport(from = initialReportPath, to = self$reportFilePath, keep = TRUE)
+        if(self$createWordReport){
+          file.copy(
+            from = gsub(pattern = ".md", replacement = ".docx", x = initialReportPath), 
+            to = gsub(pattern = ".md", replacement = ".docx", x = self$reportFilePath)
+            )
+        }
```
Yuri05, line 121: this code is exactly the same for all 3 workflow types. 
It would be better if e.g. the function `copyReport `would have additional argument (`copyWordReport `or so) and lines 117-120 would be moved into the copyReport function.
Then you would replace lines 115-121 with
`copyReport(from = initialReportPath, to = self$reportFilePath, keep = TRUE, copyWordReport =self$createWordReport)`
same for meanmodel and pop-workflow

Yuri05:
Looks good. merging

PR #820, opened 2022-06-02
817 subunits alphabetical

pchelle:


R/qualification-ddi.R
```diff
@@ -242,43 +242,38 @@ buildQualificationDDIDataframe <- function(dataframe,
...
+    minLimit <- ifelse(
+      min(dataVector)>=1,
+      min(0.8*min(dataVector), 0.8),
+      min(0.8*min(dataVector), 0.25)
+    )
+    maxLimit <- ifelse(
+      max(dataVector)>=1,
+      max(1.25*max(dataVector), 4),
+      max(1.25*max(dataVector), 1.25)
+    )
```
Yuri05, line 261: Why do we need the lines 249-261?
(minLimit, maxLimit) returned here is the same as in the lines 265-276 below.

pchelle: Actually we don't !!
The code was mirroring the Matlab code, which was formulated this way because the function was using user-defined limits instead of min/max of data.

PR #825, opened 2022-06-21
Fixes #821 prevent too long legends to be truncated

### LINKED ISSUE #821: DDI Ratio plots: The (large) legend is truncated.
Example from here: https://github.com/Yuri05/TestReport/blob/main/Report03_DDI_CYP3A4/report.md#1225

image

pchelle:
The fix update the final plot size and use the


pchelle:
Here is the updated DDI Plot
DDIRatio_1_ddi_ratio_plot_AUC_predictedVsObserved

PR #829, opened 2022-06-23
826 ddi default shapes

pchelle:


Yuri05:
5 tests fail now...

pchelle:
This is an interesting error for when you search an element from a list
```r
GOFMergedPlot <- list(PlotTypes = "observedVsPredicted")
```
Then, if you write
```r
GOFMergedPlot$Plot 
> "observedVsPredicted"
GOFMergedPlot$PlotTypes
> "observedVsPredicted"
```
but if you write
```r
GOFMergedPlot[["Plot"]]
> NULL
GOFMergedPlot[["PlotTypes"]]
> "observedVsPredicted"

```

So basically, using `$` operator, R gets an element that partially match elements from lists (which caused a problem here)

PR #839, opened 2022-06-28
Fixes #837 enforce removal of Caption in legend title

### LINKED ISSUE #837: DDI Ratio plots: remove "Caption" from the legend
grafik

pchelle:
@Yuri05 I did not get this issue when testing with my current R version, could you check that this fix actually solved the issue


Yuri05:
interesting.
Yes, fixed the issue on my comp!

Yuri05:
@pchelle 
do we need to fix this issue for any other plot types (PK-ratio etc.)?

pchelle:
The remaining plots should be okay, the issue came from `ggplot2::guide_legend` used in the ddi ratio plots that enforces the legend to stay in one unique column, however its side effect reset the title of the legend

Yuri05:
ok, cool, thx!

PR #905, opened 2022-08-05
Fixes #548 Fixes #268 refactor logging

### LINKED ISSUE #268: Improve logging and error message handling
Most of the logErrorThenStop calls do not provide the logFolder so they end up in the current directory usually duplicating the logs
I am thinking of 2 ways to improve this:
- Should we provide the logFolder to the checking method and if left empty just throw the error and warning without saving within the log ?
- Should we encapsulate the checking method as an input expression of logErrorThenStop (the expression is evaluated within the function and then logged) ?
A sys.calls() method could also be included in such a function so the error message can provide the RE function responsible for the error or warning

### LINKED ISSUE #548: Make logging more transparent

pchelle:
Refactoring logging made a lot of changes throughout the code,
there may be some glitches and bugs to fix due to this update

Now the good sides of the update:
- reEnv includes a `log` field that manage the logging system centrally
   - no need to use logFolder anymore
   - logFolder, logFolderPath... were all removed from the functions
   - log messages can be accessed using `showLogMessages` and can be re-saved using `saveLogs` and `saveLogsToJson`
   - the function `displayDebugLogs()` allows user to display all the debug messages on their R/RStudio console
- Most of the messages were moved to messages.R where I initiated some sections in the code to split between errors, warnings, info and debug messages
- I added some new features that I wanted to try
   - loading and running of simulations now display an actual progress bar
   - the message printed in the console uses the `{crayon}` package if installed to display the messages in color and highlighting keywords


Yuri05:
warning during package installation: 
`Note: possible error in 'logDebug(message = paste0(ifNotNull(nodeName, ': unused argument (file = debugLogFileName)`

PR #921, opened 2022-08-22
920 wordlist

Yuri05:


.github/workflows/testRE.R
```diff
@@ -1,3 +1,4 @@
+# spelling::spell_check_package()
```
Yuri05, line 1: I tried to include it into Linux builds, but somehow I could not see the output of the spellchecker.
So I added it to the windows build (appveyor.yml) instead

inst/WORDLIST
```diff
@@ -0,0 +1,575 @@
...
+XYGDataMapping
+yaml
+ylabel
+yLabel
+ymax
+ymin
+yParameter
+yParameters
+yticklabels
+yticks
```
Yuri05, line 575: valid words which are not recognized by the spellchecker otherwise

PR #935, opened 2022-09-28
888 gof plots

pchelle:


Yuri05:
Build failed
```
Error: processing vignette 'goodness-of-fit.Rmd' failed with diagnostics:
Error in data.frame(names = colorBreaks, labels = colorBreaks, values = finalColorValues, : arguments imply differing number of rows: 0, 1
```

pchelle:
> Build failed
> 
> ```
> Error: processing vignette 'goodness-of-fit.Rmd' failed with diagnostics:
> Error in data.frame(names = colorBreaks, labels = colorBreaks, values = finalColorValues, : arguments imply differing number of rows: 0, 1
> ```

I am trying to figure out why but can't reproduce locally.
The error seems to happen when calling the function `plotTimeProfile` from `tlf` but when I use the last .zip artifact, I don't get the error.

pchelle:
Now it fails at when calling ggplot2 manual scale... but not for every job.
`! Insufficient values in manual scale. 2 needed but only 1 provided`
So, the data might be the same between the jobs either due to different behaviors from the data mapping objects or from the building of the initial dataset

PR #939, opened 2022-10-13
888 residuals

pchelle:


Yuri05:
some tests failed on all systems...

NAMESPACE
```diff
@@ -1,5 +1,7 @@
 # Generated by roxygen2: do not edit by hand
 
+export("geomean*sd")
+export("geomean/sd")
```
msevestre, line 4: Is this a valid function name

Yuri05: @pchelle Please rename the functions. You have to call the functions in quotes otherwise 
```
`geomean*sd`(c(1,2,3))
```
Not sure if export must contain the quotes as well then, but having such a function name is pretty crazy from my point of view.

msevestre: I agree. I don't even know why this is allowed

R/reportingengine-env.R
```diff
@@ -250,3 +249,135 @@ setDefaultAutoAxisLimitMargin <- function(margin) {
...
+    ))
+  }
+  #TODO : define geometric mean in tlf !
+  return(list(
+    y = "geomean",
+    ymin = "geomeanDividedBySD",
+    ymax = "geomeanMultipliedBySD",
+    yCaption = "geometric mean",
+    # The unicode character below is supposed to be */ symbol
+    rangeCaption = "mean \u22c7 geometric SD range"
```
Yuri05, line 335: was the caption supposed to be "mean */ geometric SD range"?
because when entering this in R I get:
```
> "mean \u22c7 geometric SD range"
[1] "mean ⋇ geometric SD range"
```

R/reportingengine-env.R
```diff
@@ -250,3 +249,135 @@ setDefaultAutoAxisLimitMargin <- function(margin) {
...
+getStatisticsFromType <- function(statisticsType){
+  validateIsIncluded(statisticsType, StatisticsTypes)
+  if(isIncluded(statisticsType, StatisticsTypes$`2.5th-97.5th Percentiles`)){
+    return(list(
+      y = tlf::tlfStatFunctions$`Percentile50%`,
+      ymin = tlf::tlfStatFunctions$`Percentile2.5%`,
+      ymax = tlf::tlfStatFunctions$`Percentile97.5%`,
+      yCaption = "median",
+      # The unicode characters below are superscript th
+      rangeCaption = "[2.5\u1d57\u02b0-97.5\u1d57\u02b0] percentiles"
```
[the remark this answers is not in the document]
pchelle: On my local computer it does.
I checked the default printing documentation that indicates the following:
_Whether a character is printable depends on the current locale and the operating system_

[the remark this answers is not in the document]
pchelle: And when checking using vscode on the md document, it also does look good.
My plots also looks okay, but I don't know if they will on your side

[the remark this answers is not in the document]
Yuri05: interesting... 
OK, let's keep it for now AS IS, I will check later how the reports look like e.g. on my machine

R/reportingengine-env.R
Yuri05: Let's use "*/".
After looking at the symbol in very big resolution (https://www.compart.com/en/unicode/U+22c7) I finally understood what it is, but in the smaller size it just looks like "*"

R/utilities-goodness-of-fit.R
```diff
@@ -553,6 +572,65 @@ plotMeanTimeProfile <- function(simulatedData,
...
+                                   observedData = NULL,
+                                   lloqData = NULL,
+                                   metaData = NULL,
+                                   dataMapping = NULL,
+                                   plotConfiguration = NULL) {
+  # Remove 0 values from simulated and observed data
+  # TODO: use dplyr to refactor the selection
+  simulatedData <- removeNegativeValues(simulatedData, dataMapping$y)
+  observedData <- removeNegativeValues(observedData, dataMapping$y)
+  lloqData <- removeNegativeValues(lloqData, dataMapping$y)
```
Yuri05, line 596: There are some problems with this solution. 
**But I would suggest to address them in the separate PR in order not to block this PR for even longer.**
(just created an issue for this: https://github.com/Open-Systems-Pharmacology/OSPSuite.ReportingEngine/issues/942)

I have already created the issue for this in the ospuite-R (because residuals are calculated there as well):
https://github.com/Open-Systems-Pharmacology/OSPSuite-R/issues/1091

In short:
* Removing <=0 values (especially observed values) is not correct. E.g. if we have `observed = 0` and `simulated >>0`, we just ignore a significant residual in such a case.
* Very small positive values are problematic for log residuals as well. E.g. having `observed = 1e-100` and `simulated = 1e-200` would produce a huge log residual, but in fact the difference is irrelevant - both values are numerical zero.

For this, PK-Sim/MoBi introduces the `LogSafe `function, which is used for calculating log residuals and defined as:
`LogSafe(Y) = Log(Max(Y, LogSafeEpsilon))` (with LogSafeEpsilon currently defined as 1E-20)

And Lin residuals are calculated with all the values AS IS

Yuri05 (approved):

PR #944, opened 2022-10-25
881 data selection

pchelle:


R/output.R
```diff
@@ -86,16 +84,18 @@ Output <- R6::R6Class(
...
-        if (dataSelection %in% DataSelectionKeys$NONE) {
-          self$dataSelection <- NULL
+        # When concatenating, ALL won't be understood by dplyr
+        # Needs to be replaced by true to select all data
+        dataSelection[dataSelection %in% DataSelectionKeys$ALL] <- TRUE
+        # Concatenate selections using &
+        self$dataSelection <- paste(dataSelection, collapse = " & ")
+        # If any selection include None, do not select anything
+        if (isIncluded(DataSelectionKeys$NONE, dataSelection)) {
+          self$dataSelection <- FALSE
```
Yuri05, line 98: brackets are missing when concatenating with &.

E.g. consider the case `dataSelection=c('2<3 | 4<3 ', '5<3')` 
The expectation is FALSE, because the last term `'5<3'` is FALSE. But:
```
> paste(dataSelection, collapse = " & ")
[1] "2<3 | 4<3  & 5<3"
> 2<3 | 4<3  & 5<3
[1] TRUE
```

So if the brackets are not somehow added before the function call: they must be added here

R/simulation-set.R
```diff
@@ -75,6 +82,20 @@ SimulationSet <- R6::R6Class(
...
+        # When concatenating, ALL won't be understood by dplyr
+        # Needs to be replaced by true to select all data
+        dataSelection[dataSelection %in% DataSelectionKeys$ALL] <- TRUE
+        # Concatenate selections using &
+        self$dataSelection <- paste(dataSelection, collapse = " & ")
+        # If any selection include None, do not select anything
+        if (isIncluded(DataSelectionKeys$NONE, dataSelection)) {
+          self$dataSelection <- FALSE
+        }
+      }
```
Yuri05, line 98: code duplication with the output class. consider definition as a function

R/utilities-observed-data.R
```diff
@@ -95,6 +195,7 @@ readObservedDataFile <- function(fileName,
 #' @return vector of logicals corresponding to the evaluation of the filter
 #' @export
 evalDataFilter <- function(data, filterExpression) {
+  .Deprecated("getSelectedRows")
```
Yuri05, line 198: nice

R/utilities-observed-data.R
```diff
@@ -85,6 +85,106 @@ readObservedDataFile <- function(fileName,
...
+    return(data)
+  }
+  if(isIncluded(dataSelection, DataSelectionKeys$NONE)){
+    return(data[FALSE,])
+  }
+  if(isOfType(dataSelection, "expression")){
+    return(data %>% dplyr::filter(eval(dataSelection)))
+  }
+  return(data %>% dplyr::filter(eval(parse(text = dataSelection))))
+}
```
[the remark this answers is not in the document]
pchelle: Interesting, I did not think about that issue...
`isEmpty` checks the length of the object, I did not expect user input `dataSelection=""`.
In this case, I would provide the same behaviour as `dataSelection=NULL`.
Note that `dataSelection=NA` would select no data, because of the behaviour of the function `dplyr::filter`

```r
> length("")
[1] 1

> as.character(NULL)
character(0)

> isEmpty(as.character(NULL))
[1] TRUE

> is.null(as.character(NULL))
[1] FALSE
```

R/utilities-observed-data.R
```diff
@@ -407,3 +504,30 @@ getObservedDataIdFromPath <- function(path) {
...
+    return(dataSelection)
+  }
+  # If any selection include None, do not select anything
+  if (isIncluded(DataSelectionKeys$NONE, dataSelection)) {
+    return(FALSE)
+  }
+  # By removing "" string from dataSelection
+  # If "" is the only value provided, dataSelection isEmpty
+  # If multiple values provided, concatenate the remaining selections
+  dataSelection <- dataSelection[!(dataSelection %in% "")]
```
Yuri05, line 526: before that, dataSelection should be trimmed, e.g. with str_trim() from stringr:
```
> dataSelection=c(" aa ", " bb ", " ")
> dataSelection=str_trim(dataSelection)
> dataSelection
[1] "aa" "bb" ""  
```

Yuri05: After the current line 526 we should also add the check `if(isEmpty(dataSelection)) return(FALSE)`
Otherwise the function returns "()" which is not really convenient and also not consistent with the FALSE return in case of `DataSelectionKeys$NONE`
(and then you could also remove checking for "()" when calling the function ;)

R/utilities-observed-data.R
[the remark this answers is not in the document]
Yuri05: ok, I created the issue for the isEmpty function in the ospsuite.utils

Yuri05 (approved):

PR #947, opened 2022-11-07
Fixes #872 Fixes #883 introduce Output properties groupID, color, fill

### LINKED ISSUE #872: Output: groupID
In MeanModel/Population workflows in the Time Profiles: all outputs with the same display unit are currently placed into the same plot.
Sometimes users want however to plot those outputs in **separate** plots.

For this:
* add new optional property **GroupID** (string) to the Output class
* when creating time profile plots: split outputs by the combination {Display Unit; GroupID}. Each combination becomes then a separate plot.

### LINKED ISSUE #883: MeanModel/Population-Workflow: unique colouring schema for each Output
After #881 and #882 are implemented: assign unique colour schema
* range color **(better name for that?)** - used for 
  * the selected range in the pop simulations, depending on the selected statistics type.
  * histograms (e.g. distribution of residuals)
  * ... (anything else?)
* line colour **(better name for that?)** - used for 
  * the simulated curve (with the new spec we will have exactly 1 curve for each output also in the pop sims)
  * observed data assigned to this output

Exception: in population time profile plots with 2 populations (nonref-pop and ref-pop) the reference population (both simulated and observed data) should be displayed in another colors (e.g. the "greyscale version" of the corresponding output colours ?)

pchelle:


pchelle:
I think the package `{lifecycle}` is responsible for the failures on the Linux tests.
In version [1.0.3](https://lifecycle.r-lib.org/news/index.html) soft deprecation is now warned (mass balance and demography plots used soft deprecated ggplot2 functions).
The warning is caught in the error log file and lead to the test failure counting a mismatch in the log files.

PR #956, opened 2022-11-21
882 excel template

pchelle:


Yuri05:
@pchelle Very nice.
One remark on the excel template: can you extend the explanations for colours, e.g. that "color" is for lines and points(?) and "fill" for ranges.
grafik

PR #958, opened 2022-12-01
858 pk parameters

pchelle:


Yuri05:
some tests fail on Windows

PR #962, opened 2022-12-09
919 invalid links

pchelle:


R/utilities-configuration-plan.R
```diff
@@ -1,4 +1,4 @@
-# Helper functions to translate configuration plan values/fields to tlf/opsuite.reportingengine
+# Helper functions to translate configuration plan values/fields to tlf/ospuite.reportingengine
```
Yuri05, line 1: closer, but still a typo :)
**ospsuite**

PR #964, opened 2022-12-20
Fixes #849 watermark is rescaled if bigger than plot dimensions

### LINKED ISSUE #849: Scaling of watermark: should fit plotting area
grafik

pchelle:


R/utilities-plots.R
```diff
@@ -350,6 +350,49 @@ prettyCaption <- function(captions, plotObject, element = "legend") {
...
+  # No need to update if no displayed watermark
+  if (isEmpty(plotObject$plotConfiguration$background$watermark$text)) {
+    return(plotObject)
+  }
+  # Watermark size in inches to compare with plot dimensions
+  # Font size is in point = 1/72 inches
+  watermarkSize <- nchar(plotObject$plotConfiguration$background$watermark$text) *
+    plotObject$plotConfiguration$background$watermark$font$size / 72
+  watermarkWidth <- abs(watermarkSize * cos(plotObject$plotConfiguration$background$watermark$font$angle * pi / 180))
+  watermarkHeight <- abs(watermarkSize * sin(plotObject$plotConfiguration$background$watermark$font$angle * pi / 180))
```
Yuri05, line 368: Why `*pi/180` ?

pchelle: Because the angle from `font` is in degrees and the `cos`/`sin` functions use radians.
So it needs to be converted to get the correct width/height if there is an angle in the watermark.
For the conversion,  $360^o = 2 \pi$

PR #975, opened 2023-01-23
Fixes #968 use writeLines for special characters

### LINKED ISSUE #968: Special characters not properly displayed in the report
S. the reports here: https://github.com/Yuri05/TestReport/tree/main/Report_07_Population_Pediatric_Raltegravir (Report inputs are in the same folder)

I tried it with R3.6 and R4.1.
In both cases, the unicode chars which should represent <sup>th</sup> are not properly displayed (both in md and word report)

e.g. https://github.com/Yuri05/TestReport/blob/main/Report_07_Population_Pediatric_Raltegravir/Report_R413/Report.md#211-c_max
grafik

pchelle:
@Yuri05 could you let me know if the fix works on non-utf environment ?

### DIFF
--- a/R/utilities-writing-report.R
+++ b/R/utilities-writing-report.R
@@ -44,7 +44,7 @@ addFigureChunk <- function(fileName,
   )
 
   fileObject <- file(fileName, encoding = "UTF-8", open = "at")
-  write(mdText, file = fileObject, append = TRUE, sep = "\n")
+  writeLines(mdText, con = fileObject, sep = "\n", useBytes = TRUE)
   close(fileObject)
 
   usedFilesFileName <- sub(pattern = ".md", replacement = "-usedFiles.txt", fileName)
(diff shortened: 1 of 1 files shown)


Yuri05:
@pchelle Well, not really fixed...
grafik

Yuri05:
`<sup>th</sup>` does not work?
At least within the GitHub markdown it works: **5<sup>th</sup> percentile**
Or the word conversion does not look good then?

Yuri05:
@pchelle OK, so the following works for me:
* Markdown: `<sup>th</sup>`
* Word: replace `<sup>th</sup>` with `^th^` and use the [supperscript extension](https://pandoc.org/MANUAL.html#superscripts-and-subscripts) during the conversion

E.g. if my `report_word.md` looks like this:
```
# Lalala

5^th^ percentile of x = Y
```

then I convert via `pandoc report_word.md -o report.docx -f markdown+superscript`
and the word report looks as expected:
grafik

PR #984, opened 2023-01-31
536 monte carl

pchelle:


R/utilities-plots.R
```diff
@@ -642,6 +642,59 @@ getGOFPlotConfiguration <- function(plotType,
...
+#' @keywords internal
+getBoxWhiskerPlotConfiguration <- function(plotScale = "log",
+                                           colorGrouping = NULL,
+                                           data,
+                                           metaData,
+                                           dataMapping = NULL,
+                                           plotConfiguration = NULL) {
+  if (!isEmpty(plotConfiguration)) {
+    return(plotConfiguration)
+  }
```
Yuri05, line 662: I see a problem wtih this implementation, because it makes it virtually impossible for users to make adjustions to the default BoxWhisker plot configuration.
But it seems to me a bigger topic (which might be also relevant for other kinds of plots), so I will create an Issue for this on GitHub, which can be adressed in further PRs

Yuri05 (approved):

Yuri05:
Merging for now. Will create an issue for the settings, which can be addresed in another PR

PR #986, opened 2023-02-03
Fixes #923 word report includes bookmarks for figures and tables

### LINKED ISSUE #923: Text bookmarks not created in word for the anchors from markdown
I did not find the way to enforce the bookmarks generation.
As a workaround:
* convert to **html** first: pandoc -o ".../report-word.html" ".../report-word.md"
* then convert **from html** to docx using the same settings as now

As I am not sure about potential side effects I would use the workaround above for the **_qualification_** reports only.

E.g.: direct conversion md=>docx produces bookmarks for the md-headers only.
Conversion via html also adds all figures and tables.
MD=>DOCX       ➡️ MD=>HTML=>DOCX

pchelle:
Note that with this new method, page breaks don't require lua filters any longer


R/utilities-writing-report.R
```diff
@@ -231,10 +231,26 @@ renderWordReport <- function(fileName, intro = NULL, createWordReport = FALSE, w
...
+        '<w:br w:type="page"/>',
+        "</w:r><w:r>",
+        # Bookmark that uses anchore name (e.g. figure-1-1)
+        '<w:bookmarkStart w:id="', anchorName, '" w:name="', anchorName, '"/>',
+        '<w:bookmarkEnd w:id="', anchorName, '"/>',
+        "</w:r></w:p>"
+        ),
+      "```",
+      ""
+      )
```
Yuri05, line 249: @pchelle Can you add the link to the webpage with this hint as a comment?

pchelle: In case, also adding the link as an answer here
https://learn.microsoft.com/en-us/dotnet/api/documentformat.openxml.wordprocessing.bookmarkstart?view=openxml-2.8.1#remarks

Yuri05 (approved):

PR #990, opened 2023-02-08
Fixes #951 prevent truncated axis ticklabels

### LINKED ISSUE #951: Qualification report: GOF plot and concentration-time plots truncated
The attached evaluation plan (zip-file) produces a GOF plot and concentration-time profile plots which are truncated so that 1500 appears as 150.

image

image

[7E3-Model__Kopie.zip](https://github.com/Open-Systems-Pharmacology/OSPSuite.ReportingEngine/files/10033573/7E3-Model__Kopie.zip)

pchelle:
On a few tested plots, the value of 20 pts seems the best compromise to prevent truncated ticklabels without changing the final shape of the plot. If necessary, an optimal value can be calculated as a ratio of the final size of the plot


Yuri05:
ok, let's merge for now and see how it work

PR #993, opened 2023-02-20
Fixes #934 qualification workflows handle dual axis in time profiles

### LINKED ISSUE #934: Missing Y2 in plots of Qualification Report (Workflow.R)
I tried to generate the qualification report for Fluconazole model based on files provided here:
https://github.com/Open-Systems-Pharmacology/Fluconazole-Model

However, instead of Workflow.m I have used workflow.R as described here:
https://www.open-systems-pharmacology.org/OSPSuite.ReportingEngine/articles/qualification-workflow.html#template

I have PK-Sim portable version 11 and latest version of 'QualificationRunner'. The issue is in concentration-time profiles with double Y axes like Y and Y2 (like the third plot in section 3.3.1), I got a plot with only Y and not both Y and Y2. Can you please help me with this issue?

### LINKED ISSUE #985: Qualification workflow / Time profile plots: add 2nd y-axis automatically if required
At the moment, when running a Qualification Workflow with time profile plots coming from PK-Sim:
if a plot contains the second Y axis - it is ignored with warning:

```
! Warning	Right Axis not available yet in tlf package,
Wagner 1981 - 0.5mg po - Digoxin - PO - 0.5 mg - Urine - agg. (n=6)-Digoxin-Kidney-Urine-ArithmeticMeancould not be added
```

In such a case RE should automatically add the 2nd y-axis, without any need for the user to specify this in the configuration

Related to: https://github.com/Open-Systems-Pharmacology/TLF-Library/issues/392

pchelle:
Also Fixes #985

The fix requires PR [407](https://github.com/Open-Systems-Pharmacology/TLF-Library/pull/407) from tlf package and might have side effects on the legend order of other plots performed by the reporting engine


Yuri05:
5 tests fail on windows...

R/qualification-time-profile.R
```diff
@@ -78,144 +81,186 @@ plotQualificationTimeProfiles <- function(configurationPlan, settings) {
...
+    plotObject <- updatePlotAxes(plotObject, axesProperties)
+    return(plotObject)
+  }
+  # Y, Y2 and Y3 defined and used
+  if (isOfLength(requestedAxes, 3)) {
+    logError(messages$warningTooManyAxes(
+      project = timeProfilePlan$Project,
+      simulation = timeProfilePlan$Simulation
+    ))
+    requestedAxes <- c("Y", "Y2")
```
Yuri05, line 221: Why this check 2 times (here and above lines 189-191)?

R/qualification-time-profile.R
```diff
@@ -78,144 +81,186 @@ plotQualificationTimeProfiles <- function(configurationPlan, settings) {
...
+  }
+  # Use dual axis but needs to map them first
+  # New Y axis will be either Y or Y2 (ie. axis found first because they were sorted)
+  # New Y2 axis will be either Y2 or Y3 (ie. axis found last because they were sorted)
+  axesProperties$y <- axesProperties[[head(tolower(requestedAxes), 1)]]
+  axesProperties$y2 <- axesProperties[[tail(tolower(requestedAxes), 1)]]
+
+  simData$yAxis <- ifNotNull(simData$yAxis, simData$yAxis %in% tail(requestedAxes, 1))
+  obsData$yAxis <- ifNotNull(obsData$yAxis, obsData$yAxis %in% tail(requestedAxes, 1))
```
Yuri05, line 230: @pchelle Can you explain what happens in these 2 lines?

R/qualification-time-profile.R
pchelle: Copy/Paste typo.
I am going to remove it.

R/qualification-time-profile.R
pchelle: For the right axis feature in the `{tlf}` package, the variable mapping which axis is used needs to provide logical values such as `TRUE` correspond to values on the right axis and `FALSE` values on the left axis.

For intellisense, I am replacing yAxis by y2Axis while adding the explanations below
- `FALSE` y2Axis values correspond to data on the left axis
- `TRUE` y2Axis values correspond to data on the right axis
- Since yAxis values are included in "Y", "Y2" or "Y3", `TRUE` values will correspond to the values included (`%in%`) last found axis

PR #1002, opened 2023-03-07
Fixes #925 Rbuildignore

### LINKED ISSUE #925: Add folders to Rbuildignore
Add folders to Rbuildignore

- scripts
- .github

Yuri05:


.Rbuildignore
```diff
@@ -5,3 +5,12 @@
...
 ^\.covrignore$
+^\.github$
+^docs$
+^scripts$
+^LICENSE\.md$
+^_pkgdown\.yml$
+^rakefile\.rb
+^inst/extdata/RaltegravirSim\.pkml$
+^tests/data/input-data/RaltegravirSim\.pkml$
+^tests/data/input-data/individualPksimSim\.pkml$
```
Yuri05, line 16: Those pkmls are not used in the unit tests; only in dev-tests.
also not sure why we have RaltegravirSim.pkml twice.

The file under inst/extdata was also much bigger than under tests/data/input-data and I have replaced it with another one.

Later we should keep only one version nayways.

pchelle (approved):
Except for tracelib, I think most of dev-tests are now somewhat included in the vignettes.
I think it may even be okay to remove those dev-tests from the package.

Thoughts ?

Yuri05:
> Except for tracelib, I think most of dev-tests are now somewhat included in the vignettes. I think it may even be okay to remove those dev-tests from the package.
> 
> Thoughts ?

For me OK to remove. 
@msevestre Any concenrs?

PR #1018, opened 2023-03-22
Fixes #1011 Increment version 2.1=>2.2

### LINKED ISSUE #1011: Increment version 2.1=>2.2

### LINKED ISSUE #1012: Exclude PRs when installing packages from the latest AppVeyour build
compare 
https://github.com/Open-Systems-Pharmacology/TLF-Library/blob/ee8ac83a17c13fbd40c37d4273ba740bf62d5245/appveyor.yml#L55

Yuri05:
Fixes #1012 Exclude PRs when installing packages from the latest AppVeyour build

### DIFF
--- a/.github/workflows/Check_utf8.yml
+++ b/.github/workflows/Check_utf8.yml
@@ -1,0 +1,13 @@
+name: Check if all required files are UTF-8
+
+on: 
+  [push, pull_request]
+
+permissions:
+  contents: read
+
+jobs:
+  utf8check:
+    uses: Open-Systems-Pharmacology/Workflows/.github/workflows/UTF8Check.yml@main
+    with:
+      file-extensions: 'R\|Rmd\|md'  #File extension(s) to check separated by \| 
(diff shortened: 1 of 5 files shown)


.github/workflows/Check_utf8.yml
```diff
@@ -0,0 +1,13 @@
...
+  [push, pull_request]
+
+permissions:
+  contents: read
+
+jobs:
+  utf8check:
+    uses: Open-Systems-Pharmacology/Workflows/.github/workflows/UTF8Check.yml@main
+    with:
+      file-extensions: 'R\|Rmd\|md'  #File extension(s) to check separated by \|
```
Yuri05, line 13: Check that all R, Rmd and md files are saved as UTF-8

.github/workflows/Dockerfile
Yuri05, line 1: tried to integrate lintr into the workflow, but somehow it does not work
(lintr starts and produces no results).
So in fact no changes here - just wanted to docuemnt what was tried.
There are different ready-to-use lint GitHub-Actions available - those might run better

.github/workflows/install_packages_cran.sh
Yuri05, line 1: same here - no changes, just for the documentation.
Interestingly, when trying to install lintr without explicit version (`install.packages("lintr")`)- it only installs a much older version (2.x).
R is amazing :)

PR #1021, opened 2023-03-23
Fixes #949 add new symbols to dictionary

### LINKED ISSUE #949: Add new symbols in dictionary
[New available symbols](https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/issues/1789) need to be included in dictionary

pchelle:
Note that the new method leads to smaller symbols that are scaled bigger in the RE theme

### DIFF
--- a/R/utilities-configuration-plan.R
+++ b/R/utilities-configuration-plan.R
@@ -36,7 +36,6 @@ ConfigurationShapes <- list(
   circle = tlf::Shapes$circle,
   cross = tlf::Shapes$cross,
   diamond = tlf::Shapes$diamond,
-  dot = tlf::Shapes$circle,
   hexagon = tlf::Shapes$hexagon,
   invertedtriangle = tlf::Shapes$invertedTriangle,
   pentagon = tlf::Shapes$pentagon,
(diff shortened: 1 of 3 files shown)


R/utilities-configuration-plan.R
```diff
@@ -26,21 +26,38 @@ ConfigurationScales <- list(
...
   none = tlf::Shapes$blank,
-  circle = tlf::Shapes$hollowCircle,
-  dot = tlf::Shapes$dot,
-  square = tlf::Shapes$hollowSquare,
-  diamond = tlf::Shapes$hollowDiamond,
-  asterisk = tlf::Shapes$asterisk,
+  circle = tlf::Shapes$circle,
   cross = tlf::Shapes$cross,
+  diamond = tlf::Shapes$diamond,
+  dot = tlf::Shapes$circle,
```
Yuri05, line 39: let's remove "dot".
It is not available in PK-Sim; in case it was used in a qualification plan - we can change it to another symbol there.

Yuri05: (I don't think "dot" was ever used in a quali plan anyways)

Yuri05 (approved):

PR #1026, opened 2023-03-27
1024 text size

pchelle:


Yuri05:
better. The only thing is now that the axis captions are slightly cut off


However when trying the examples from 
https://www.open-systems-pharmacology.org/OSPSuite-R/articles/data-combined-plotting.html



With tlf 1.5.130:


So something needs to be done in the tlf as well.

pchelle:
Indeed, TLF should ensure that the font and symbol sizes are dpi independent.
> The graphic device in R Studio is 96 dpi while plots are saved in 300 dpi.

pchelle:
A temporary RE solution is to revert to 96 dpi after the plot has been saved.

Yuri05:
ok, with this PR and using the TLF from the PR https://github.com/Open-Systems-Pharmacology/TLF-Library/pull/444 the plots look good again (only the symbol size in the time profile plots become a bit smaller).
You can compare the plots here https://github.com/Yuri05/Digoxin-Model/pull/1/files

Merging this now.

PR #1031, opened 2023-04-06
Fixes #932 error and warning better handled in PK Ratio task

### LINKED ISSUE #932: Qualification workflow / PK-Ratio plots: better error handling if PK-Parameter is not available for simulation
[4mail_2.zip](https://github.com/Open-Systems-Pharmacology/OSPSuite.ReportingEngine/files/9616955/4mail_2.zip)
In the workflow attached, RE crashes with the error message:

```
i Info	Starting run of Plot PK Ratio task
Error: Error in validateIsOfType(quantityOrDimension, c("Quantity", "character")): source: argument 'quantityOrDimension' is of type 'NULL', but expected 'Quantity, or character'!
```

what happens here: PK-Ratio should be plotted for the PK-Parameter AUC which is not available for the (multiple dosing) simulation.

So actually line 254 should return NULL or so, RE should check for that and continue with the next PK-Parameter (**with Warning "PK-Parameter AUC not found for the simulation XYZ in PK-Ratio Plot X, Group Y, Ratio Z"** or so)

Instead, some kind of dummy PK-Parameter is returned and RE crashes later during unit conversion because `pkParameter$dimension` is NULL in the line 263

https://github.com/Open-Systems-Pharmacology/OSPSuite.ReportingEngine/blob/f50345c4cc0e533ec0973ada66a9f0e6ef8b0cee/R/qualification-pk-ratio.R#L253-L267

pchelle:


Yuri05:
@pchelle Please add tests

Yuri05:
new test failed...

PR #1040, opened 2023-05-02
Fixes #789 include delta parameter from guest et al formula

### LINKED ISSUE #789: DDIRatioPlot: Add delta parameter to Guest criterion
Add delta parameter to Guest criterion and make it available in to be specified in the plot settings of the qualification plans.

The parameter delta is from the interval [1, 2].

Here is the definition from the publication (https://dmd.aspetjournals.org/content/39/2/170.short):
image

This limit is then used to calculate the upper and lower limit (as before):
image

pchelle:


R/qualification-ddi.R
```diff
@@ -457,7 +454,10 @@ getDDISection <- function(dataframe, metadata, sectionID, idPrefix, captionSuffi
       id = tableID,
       sectionId = sectionID,
       table = ddiTableList[[pkParameter]],
-      tableCaption = paste("Summary table for", metadata$title, "-", pkParameter, "Ratio"),
+      tableCaption = paste(
+        "Summary table for", metadata$title, "-", pkParameter, "Ratio.",
+        "(\u03b4 =", metadata$guestDelta[[pkParameter]], "in Guest *et al.* formula)"
+      ),
```
Yuri05, line 460: some blanks are missing: 
* after **"Summary table for"** (unless the title starts with a blank)
* before and after **"Ratio."**
* after **"(\u03b4 ="**
* before **"in Guest *et al.* formula)"**

Yuri05: yes, it does. closing the comment :)

R/qualification-ddi.R
```diff
@@ -634,3 +634,33 @@ ddiPKRatioColumnName <- list(
...
+      ddiRatioPlan$GuestDelta
+    })))
+  }
+  # Otherwise, update defined values
+  for (guestPlan in ddiRatioPlan$GuestDelta) {
+    for (pkParameter in guestPlan$PKParameters) {
+      guestDelta[[pkParameter]] <- guestPlan$Value
+    }
+  }
+  return(guestDelta)
```
Yuri05, line 665: This should throw an error if `pkParameter `is not in the list `ddiRatioPlan$PKParameters `

Yuri05: Maybe I'll find the way to catch this on the Qualification Plan Schema level, but I don't think it's possible there.

tests/testthat/test-ddi-ratio.R
```diff
@@ -33,3 +35,83 @@ test_that("'getSmartZoomLimits' always includes the 2-fold error range for resid
...
+  ddiRatioPlan5 <- list(PKParameters = c("AUC", "CL","CMAX"))
+  
+  expect_equal(
+    getGuestDeltaFromConfigurationPlan(ddiRatioPlan1),
+    list(AUC = 1.2, CL = 1.2, CMAX = 1.3)
+  )
+  expect_equal(
+    getGuestDeltaFromConfigurationPlan(ddiRatioPlan2),
+    list(AUC = 1.2, CL = 1.2, CMAX = 1.3)
+  )
```
Yuri05, line 72: should throw an error (s. above)

Yuri05:
both Win and Linux builds failed...

PR #1054, opened 2023-05-30
1049 delta guest

pchelle:


R/qualification-ddi.R
```diff
@@ -290,7 +290,9 @@ generateDDIQualificationDDIPlot <- function(ddiPlotData, delta = 1) {
     color = "Caption",
     minRange = c(0.1, 10),
     residualsVsObserved = residualsVsObserved,
-    deltaGuest = delta
+    # Undefined delta in Configuration Plan corresponds to NULL
+    # In such case, default value should be 1 to prevent crash when defining data mapping
+    deltaGuest = delta %||% 1
```
Yuri05, line 295: @pchelle The difference is when explicitly passing NULL as argument, right?
Before `generateDDIQualificationDDIPlot (ddiPlotData, NULL) `would result in `deltaGuest = NULL` and now it results in `deltaGuest = 1`, correct?

tests/testthat/test-ddi-ratio.R
```diff
@@ -91,7 +91,7 @@ ddiTestData <- data.frame(
 ddiTestData$simulatedObservedRatio <- ddiTestData$simulatedRatio/ddiTestData$observedRatio
 
 test_that("Guest et al. measure works as expected", {
-  ddiMeasure <- getQualificationDDIRatioMeasure(ddiTestData, pkParameterName = "AUC")
+  ddiMeasure <- getQualificationDDIRatioMeasure(ddiTestData, pkParameterName = "AUC", delta = 1)
```
Yuri05, line 94: please also add the call with `delta = NULL`

R/qualification-ddi.R
pchelle: This is correct.

PR #1074, opened 2023-06-27
Fixes #1071 auto-numbering does not mess up static images

### LINKED ISSUE #1071: "Static" images not shown
In the DDI-CYP3A4 report, all static images in the chapters 1, 5 are not displayed
https://github.com/Yuri05/TestReport/blob/main/Report_11_DDI_CYP3A4/report.md#5-appendix

grafik

pchelle:
The auto-numbering algorithm for Figures assumes that the caption is below the figure. In the qualification example, it was above. The fix renders a more robust way of including the bookmarks before the figure without needing a placeholder for the md figure inclusion line

### DIFF
--- a/tests/data/utilities-report/testReportNumFigs.md
+++ b/tests/data/utilities-report/testReportNumFigs.md
@@ -7,15 +7,9 @@
 ## Sub title 1
 
 
-<a id="figure-1-1"></a>
-
-
 Figure 1-1: this is figure 1
 
 ## Sub title 2
-
-
-<a id="figure-1-2"></a>
 
 
 Figure 1-2: this is figure 2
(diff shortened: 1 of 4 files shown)


tests/data/utilities-report/testReportNumFigs.md
Yuri05, line 1: @pchelle I wonder why in this and other test files some (all?) figure anchors have disappeared?

pchelle: It is because figure caption is below.
If I identify based on the pattern **`Figure`**, the anchor/bookmark arrives after the actual figure. 
Thus, if you click on a link to that anchor, the viewer jump at the caption location and you do not see the figure which is above.

To prevent that, 

- either you need a placeholder to memorize there was a figure.<br>
But this was not robust if users place figure captions as in the example of this issue and the figure was lost.
- or you need to identify directly the figure itself **``** which is done here<br>
Since there was no figure path in the test, no anchor was created

PR #1076, opened 2023-07-04
Fixes #536 pk parameter use individual ratios if same population

### LINKED ISSUE #536: SD and geoSD of ratio of PK parameters for population simulations seem to be wrongly calculated
The ratio of PK parameters for population workflows seems to be calculated as the ratio of the respective summary statistics of the population of interest and the control population, e.g. if the case example compares a pediatric population to an adult population the ratio summary parameters are calculated as:

mean of ratio = mean_pediatric/mean_adult
geomean of ratio = geomean_pediatric/geomean_adult
SD of ratio = sd_pediatric/sd_adult
geoSD of ratio = geosd_pediatric/geosd_adult

This calculation of ratios is an appropriate approximation in case of mean and geomean but wrong for SD and geoSD. Here, the delta method is needed. 

For the arithmetic SD this would be:
z = x/y
dzdx = 1/meanY
dzdy = -meanX/meanY^2 
meanZ = meanX/meanY
sdZ = sqrt(dzdx^2 *times* sdX^2+dzdy^2 *times* sdY^2)

For the geometric SD this would be:
z = x/y
geomeanZ = exp(log(geomeanX)-log(geomeanY))
geosdZ = exp(sqrt(log(geosdX)^2+log(geosdY)^2))

pchelle:


R/utilities-ratio-comparison.R
```diff
@@ -107,29 +113,36 @@ getPKParameterRatioPKTable <- function(data,
...
+      # After selecting the comparison data, only 1 level is remaining
+      # leading to empty factor levels for calculation of BoxWhisker measure
+      ratioData$simulationSetName <- factor(ratioData$simulationSetName)
       ratioData$Value <- comparisonData$Value / referenceData$Value
       ratioPKMeasure <- getPKParameterMeasure(ratioData, dataMapping)
       ratioPKTable <- rbind.data.frame(ratioPKTable, ratioPKMeasure)
       next
     }
-    logDebug(paste0(
+    logInfo(paste0(
```
pchelle, line 133: Previously the comparison between analytical solution and monte carlo simulations was silently saved in log-debug.
Now it will be displayed on the console and saved in log-info

R/utilities-ratio-comparison.R
```diff
@@ -224,7 +257,7 @@ isSamePopulationInSets <- function(structureSets) {
   isReference <- sapply(structureSets, function(set) {
     set$simulationSet$referencePopulation
   })
-  referenceSet <- structureSets[isReference]
+  referenceSet <- structureSets[isReference][[1]]
```
pchelle, line 260: This was where the issue actually was.
Since `structureSets[isReference]` is an array of lists, `referenceSet$simulationSet$populationFile` was `NULL`.
Thus logical `isSamePopulation` was always `FALSE`, which led to always perform the Monte Carlo simulation even when the population files were the same

tests/testthat/test-monte-carlo-sampling.R
```diff
@@ -43,12 +43,12 @@ test_that("Analytical Solution is close to known solution", {
   )
   
   expect_equal(
-    analyticalSolution$geoMean,
+    analyticalSolution$`geo mean`,
```
pchelle, line 46: This was the test implemented to check that the Monte Carlo gave appropriate solutions:

1) Use `rlnorm()` function to simulate 2 virtual population with **known** log normal distributions
   a) Reference population: $mean log=0$ and $sd log = 1$
   b) Compared population: $mean log=1$ and $sd log = 2$
2) Compute expected $geoMean$ and $geoSD$ of their ratio from the analytical formula
$geoMean = exp(mean log_{compared} - mean log_{reference}) = exp(1-0) = e$
$geoSD = exp(\sqrt{sd log_{compared}^2+sd log_{reference}^2}) = \frac{e^5}{2}$
3) Assess if analytical solution from simulated populations match expected
4) Assess if Monte Carlo solution from simulated populations match expected

PR #1078, opened 2023-07-07
Fixes #1075 bookmark conversion without linebreak

### LINKED ISSUE #1075: Generation of bookmarks from anchors
In a qualification report, I tried to add anchors to every publication reference (tried different ways: after the ref, before the ref - same line, before the ref: separate line)
grafik

Every approach resulted in different problems (s. generated code below):
* anchor which is added at the end of the line, is not replaced by the openxml code at all
* openxml code for a bookmark is always inserted in the newline, so that adding an anchor before the publication reference but in the same line splits the reference into 2 parts
* openxml code for a bookmark creates unnecessary newlines before and after
grafik

Resulting word document:
grafik

So I would suggest: 
1. Do not insert newlines before and after the bookmark
Current openxml code for a bookmark is embedded into `<w:p><w:r>` - it seems that the bookmarks are properly created when this is omitted (without additional lines then). Thus an openxml code for the bookmark would look like this:
`<w:bookmarkStart w:id="ref-iwamoto-2008" w:name="ref-iwamoto-2008"/><w:bookmarkEnd w:id="ref-iwamoto-2008"/>`
2. Insert the bookmark openxml with the "single line" code tag instead of "multiline" code tag - on the same line where the anchor is defined. Actually I would expect, that the anchor is **replaced** by the openxml bookmark code (or do we need the anchors in report_word.md for anything else?)

Thus e.g. for this text in report.md
```
<a id="ref-kasichayanula-2013"></a>**Kasichayanula 2013** Kasichayanula S, Liu X, Griffen SC, Lacreta FP, Boulton DW. Effects of
rifampin and mefenamic acid on the pharmacokinetics and pharmacodynamics of dapagliflozin.
Diabetes, obesity & metabolism. 2013;15(3):280-3. 
```
I would expect this in report_word.md:
```
`<w:bookmarkStart w:id="ref-kasichayanula-2013" w:name="ref-kasichayanula-2013"/><w:bookmarkEnd w:id="ref-kasichayanula-2013"/>`{=openxml}**Kasichayanula 2013** Kasichayanula S, Liu X, Griffen SC, Lacreta FP, Boulton DW. Effects of
rifampin and mefenamic acid on the pharmacokinetics and pharmacodynamics of dapagliflozin.
Diabetes, obesity & metabolism. 2013;15(3):280-3. 
```
Created word report looks then as expected and bookmarks are still working (tried it with pandoc 3.1.2)

pchelle:


R/utilities-writing-report.R
```diff
@@ -198,24 +198,23 @@ renderReport <- function(fileName, createWordReport = FALSE, numberSections = TR
...
+#' @param mdText Character strings of markdown text
+#' @return Updated md text
+#' @keywords internal
+translateForWord <- function(mdText) {
+  wordText <- NULL
+  # Page break using source code
+  # https://pandoc.org/MANUAL.html#extension-raw_attribute
+  artifactPageBreak <- c(
+    "```{=openxml}",
+    '<w:p><w:r><w:br w:type="page"/></w:r></w:p>',
```
Yuri05, line 213: Same as for the bookmarks: do we really need `<w:p><w:r>` here? 
If no: let's remove and keep only `<w:br w:type="page"/>`

inst/WORDLIST
```diff
@@ -112,6 +112,7 @@ displayUnit
 docx
 dplyr
 dv
+eg
```
Yuri05, line 115: it should be **e.g.**
let's remove **eg** from the wordlist and fix the function description instead

R/utilities-writing-report.R
pchelle: I checked, there is indeed no need for these tags

pchelle:
I think there is something about how appveyor and testthat work that cause the test in pop-pk-parameters to fail.
I am guessing that the file path for the monte carlo simulations is not the same due to the call of testthat, 
thus the csv cannot be open to save the debugging results.
I'll create a separate issue to fix that

PR #1083, opened 2023-07-19
Fixes #1082 no observed data does not crash qualification time profile

### LINKED ISSUE #1082: Qualification report: time profile plot crashes if no observed data is contained
Dear OSPSuite experts, I downloaded the files available in:
https://github.com/Open-Systems-Pharmacology/Evaluation-plan-template/tree/master 

and executed it with no error. Then, I simplified the 'evaluation_plan.json' file by setting: "GOFMergedPlots": [] and removed the corresponding part in the 'Sections'. This way in the report only 'Concentration-Time Profiles' are plotted (no GoF plots). Executing the reportingEngine worked well. Then, in PK-Sim, I loaded the 'Compound.json' and removed the observed data in the 'Sim 1 - intravenous' simulation and export the updated file as 'Compound.json' and executed the createQualificationReport. However, this time I got this error: 

! Warning	argument 1 is not a vector
Configuration Plan Information:
Project: Compound
Simulation: Sim 1 - intravenous
SectionReference: ct-profiles
Plot: list(...)
PlotNumber: 0 (json numbering starting at 0);  1 (actual plot index)

	
> Error Trace
	 ↪ createQualificationReport(qualificationRunnerFolder, pkSimPortableFolder)
	  ↪ workflow$runWorkflow()
	   ↪ self[[plotTask]]$runTask(self$configurationPlan)
	    ↪ self$getTaskResults(configurationPlan, self$settings)
	     ↪ ifNotNull(timeProfilePlan$Plot$Type, plotQualificationPopulationTimeProfile(simulationAnalysis = timeProfilePlan$Plot$Analysis, 
	      ↪ plotQualificationMeanTimeProfile(configurationPlanCurves = timeProfilePlan$Plot$Curves, 
	       ↪ updateQualificationTimeProfilePlotConfiguration(simulatedMetaData = simMetaData, 
	        ↪ order(observedMetaData$id)
Error: argument 1 is not a vector
Configuration Plan Information:
Project: Compound
Simulation: Sim 1 - intravenous
SectionReference: ct-profiles
Plot: list(...)
PlotNumber: 0 (json numbering starting at 0);  1 (actual plot index)

here is the zip file:
[[Files.zip](https://github.com/Open-Systems-Pharmacology/Forum/files/12039648/Files.zip)](https://github.com/GhazalMontaseri/Test2)

Two points: 
1) When I create a single datapoint observation at time=0 (or any time) and concentration = 0 mmol/l and add it to the simulation, it works again.
2) population simulations without observation work well, only with individual population I had the issue.

I would appreciate if you can help me with this issue?

_Originally posted by @GhazalMontaseri in https://github.com/Open-Systems-Pharmacology/Forum/discussions/1569_

pchelle:
@Yuri05 
I was not expecting the following behavior from the function `isEmpty` (from the package `{ospsuite.utils}`):
A named list with only `NULL` elements is not considered empty while a data.frame with only `NULL` columns is empty.

```r
> isEmpty(list(a=NULL))
[1] FALSE

> length(list(a=NULL))
[1] 1

> isEmpty(data.frame(a=NULL))
[1] TRUE

> length(data.frame(a=NULL))
[1] 0
```

Let me know if this requires also an update on the `{ospsuite.utils}` side ?

### DIFF
--- a/R/qualification-time-profile.R
+++ b/R/qualification-time-profile.R
@@ -783,14 +783,14 @@ updateQualificationTimeProfilePlotConfiguration <- function(simulatedMetaData = 
                                                             axesProperties = NULL,
                                                             plotConfiguration) {
   # Update plot configuration for simulated values
-  if (!isEmpty(simulatedMetaData)) {
+  if (!isEmpty(simulatedMetaData$id)) {
     sortedSimulatedMetaData <- order(simulatedMetaData$id)
     plotConfiguration$lines$color <- simulatedMetaData$color[sortedSimulatedMetaData]
     plotConfiguration$lines$linetype <- simulatedMetaData$linetype[sortedSimulatedMetaData]
     plotConfiguration$lines$size <- simulatedMetaData$size[sortedSimulatedMetaData]
   }
   # Update plot configuration for observed values
-  if (!isEmpty(observedMetaData)) {
+  if (!isEmpty(observedMetaData$id)) {
     sortedObservedMetaData <- order(observedMetaData$id)
     # offset of simulated data colors
     plotConfiguration$points$color <- c(


Yuri05:
>Let me know if this requires also an update on the {ospsuite.utils} side ? 

Let's create an issue there and discuss. There are obviously different opinions on `isEmpty `(compare e.g.  https://github.com/Open-Systems-Pharmacology/OSPSuite.RUtils/issues/121)

PR #1089, opened 2023-07-31
535 demography observed data

pchelle:


pchelle:
> review the article displayed from the vignette by {pkgown}

In case, I am adding the standalone html version of the vignette (which is less pretty): [demography.zip](https://github.com/Open-Systems-Pharmacology/OSPSuite.ReportingEngine/files/12216996/demography.zip)

PR #1090, opened 2023-08-03
1084 qualification anchors

pchelle:


R/captions.R
```diff
@@ -21,18 +21,18 @@ captions <- list(
   ),
   demography = list(
     parameterSection = function(sectionId, parameterName) {
-      c(anchor(sectionId), "", paste("##", parameterName, "distributions"))
+      paste("##", parameterName, "distributions", anchor(sectionId))
     },
     xParameterSection = function(sectionId, parameterName) {
-      c(anchor(sectionId), "", paste("## ", parameterName, "-dependence", sep = ""))
+      paste("## ", parameterName, "-dependence ", anchor(sectionId), sep = "")
```
Yuri05, line 27: let's put a blank between the title and the anchor everywhere, like it is done here

pchelle: There is already a blank which is included in the default arguments of the function `paste(..., sep = " ")` but this can be confusing as the function `paste0(..., sep = "")` uses a default with space

Yuri05 (approved):

PR #1104, opened 2023-08-16
1103 tlf breaking change

pchelle:


Yuri05:
Build failed due to the vignette issue with the title. Can be fixed by setting the title to the same value
Not sure though why we need 2 entries for the title if they must be the same.

grafik

pchelle:
@Yuri05 
Actually failing because tlf uses stringr v1.4.0 while the argument white space only appeared starting [version 1.5.0](https://github.com/tidyverse/stringr/issues/335)

pchelle:
> 

Not sure, but I think that all what is after vignette is used by pkgdown to create the documentation website.
So it is better when both entries are the same. If not it is actually only a warning

PR #1105, opened 2023-08-17
Fixes #1096 update y-label in histograms

### LINKED ISSUE #1096: Demographys plots - y label of histograms
In the documentation of the demography plots of the RE 2.2.0 (https://www.open-systems-pharmacology.org/OSPSuite.ReportingEngine/dev/articles/demography.html#demography-settings) I had a look on the plots and would recommend to change the y-labels of histograms from "Number of individuals [%]" to something like "Percentage of population [%]" for plots that display percentages and not counts.

pchelle:


R/utilities-demography.R
```diff
@@ -674,7 +674,7 @@ getComparisonHistogramResults <- function(demographyPaths,
       dataMapping = histogramMapping,
       plotConfiguration = settings$plotConfigurations[["histogram"]] %||%
         tlf::HistogramPlotConfiguration$new(
-          ylabel = "Number of individuals [%]",
+          ylabel = "Percentage of population [%]",
```
Yuri05, line 677: used 3 times - let's define a const for this

PR #1106, opened 2023-08-17
Fixes #1039 mass balance handles multiple applications and infusions

### LINKED ISSUE #1039: Mass balance in case of multiple applications
Currently the mass balance plots normalize the drug mass by the total final drug mass administered.
However, it does not feels right in case of multiple applications with drug mass as a function of time.
Thoughts ?

Below is an illustration of what this would change:

<details><summary><strong>Usual mass balance plot as Amount of drug vs time</strong></summary>

Raltegravir_100_mg_filmcoated_tablet_md-1_mass_balance_Raltegravir_100_mg_filmcoated_tablet_md

</details>

## Current plot

<details><summary><strong>Normalized mass balance plot normalized by total/final amount of drug</strong></summary>

Raltegravir_100_mg_filmcoated_tablet_md-3_mass_balance_Raltegravir_100_mg_filmcoated_tablet_md

</details>

## Suggested updated plot

<details><summary><strong>Normalized mass balance plot normalized by current amount of drug</strong></summary>

Raltegravir_100_mg_filmcoated_tablet_md-3_mass_balance_Raltegravir_100_mg_filmcoated_tablet_md

</details>

pchelle:


R/utilities-mass-balance.R
```diff
@@ -8,35 +8,38 @@
...
+  }
+  # Create data.frame of cumulative applied drug mass from applications
+  # which accounts for infusion time if not bolus nor oral
+  applicationResults <- getApplicationResults(applications)
+  
+  # Get all the molecule paths (with dimension=amount) for the compounds and their metabolites
+  allCompoundNames <- setdiff(
+    c(simulation$allFloatingMoleculeNames(), simulation$allStationaryMoleculeNames()),
+    simulation$allEndogenousStationaryMoleculeNames()
+  )
```
Yuri05, line 40: This is not good. We need a mass-balance plot **per applied compound** and also normalization by applied drug mass **per applied compound**. So something like this:
* For each compound in `relevantAppliedCompoundNames `
  1. get all aplications for this compound.
      * if there are no applications: continue with the next compound
  2. get all **relevant** molecules for the applied compound
  3. plot mass balance for the "relevant" molecules normalized by the drugmass of applications **for the current compound**

E.g. if I have 2 applied molecules in my simulation:
* **MyCompound**, applied in the Application **App1**
* **Itraconazole**, applied in the application **App2**

and the following metabolic network:
* `MyCompound➡️ MyCompound-CYP3A4-aa Metabolite`

* `Itraconazole ➡️ Hydroxy-Itraconazole➡️ Keto-Itraconazole➡️ N-desalkyl-Itraconazole➡️ N-desalkyl-Itraconazole ➡️N-desalkyl-Itraconazole-CYP3A4-xy Metabolite`

then  I would expect 2 plots:
1. **MyCompound** and **MyCompound-CYP3A4-aa Metabolite**, normalized by the drugmass of **App1**
2.  **Itraconazole** , **Hydroxy-Itraconazole**, etc. normalized by the drugmass of **App2**

Makes sense? @sfrechen @msevestre @Christoph27 
The question for me is how to get all the "relevant" compounds (e.g. starting from Itraconazole to get all the compounds in its metabolic chain).
* By getting all the compounds with the "Itraconazole" in the name? dangerous
* Better: by querying ospsuite-R=>OSPSuite.Core and analyzing the reactions network of Itraconazole

I don't think we have something like the last option at the moment. 
So my suggestion for now is:
* get the "relevant" compounds **by name**
* create an Issue for getting the compounds via reactions network once the feature is implemented

pchelle: > if I have 2 applied molecules in my simulation

What if the 2 applied molecules react between each other ?
I was expecting the global system (all compounds and their metabolites) to keep mass balance, but not necessarily each reaction chain of the system to keep mass balance.
Unless all the models do not create direct interactions between the applied molecules, thoughts ?

pchelle: > By getting all the compounds with the "Itraconazole" in the name? dangerous

Agree, it can either miss metabolites if they do not follow such naming convention or include unwanted paths that included the name of the molecule somehow.

> Better: by querying ospsuite-R=>OSPSuite.Core and analyzing the reactions network of Itraconazole

All in if there is a method available for this. I think exposing such a method may be valuable regardless of the mass balance workflow.

Yuri05: > What if the 2 applied molecules react between each other ? 

Possible, but very rare use case. Interaction between compounds (if any) is usually realized via [enzyme inhibition etc.](https://docs.open-systems-pharmacology.org/working-with-pk-sim/pk-sim-documentation/pk-sim-compounds-defining-inhibition-induction-processes)
I would argue that in case of **direct** reaction/interaction between several compounds or their metabolites, meaningful drug mass normalization **is not possible at all.**

> I was expecting the global system (all compounds and their metabolites) to keep mass balance, but not necessarily each reaction chain of the system to keep mass balance. Unless all the models do not create direct interactions between the applied molecules, thoughts ?

Does not make much sense for me, especially when it comes to the drug mass normalization.
If e.g. I have 2 compounds, which do not **directly** interact with each other - why should I normalize amounts of each compound by the total drug mass of all compounds? 
But let's wait for answers of @sfrechen @Christoph27 etc.

PR #1108, opened 2023-08-17
Fixes #1107 Package build: are the vignettes built twice?

### LINKED ISSUE #1107: Package build: are the vignettes built twice?
It seems to me that the vignettes are created twice:
* first, when they are built to create a package
grafik

* and second, when the package is checked
grafik

Because it takes quite a while to build the vignettes (more than 9 minutes currently): is the 2nd step ("checking rebuilding of vignette outputs") really required? If no: can it be omitted during the package check?

Yuri05:
Also deactivated Code Coverage

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -21,7 +21,7 @@ environment:
   _R_CHECK_FORCE_SUGGESTS_: false
   R_ARCH: x64
   R_VERSION: "4.3.1"
-  R_CHECK_ARGS: --no-build-vignettes
+  R_CHECK_ARGS: --no-build-vignettes --no-manual --as-cran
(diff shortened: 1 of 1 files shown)


appveyor.yml
```diff
@@ -41,7 +42,7 @@ build_script:
 
 on_success:
   - Rscript -e "spelling::spell_check_package()"
-  - Rscript -e "covr::codecov(function_exclusions = c('print.*', '\\.onLoad'))"
+#  - Rscript -e "covr::codecov(function_exclusions = c('print.*', '\\.onLoad'))"
```
Yuri05, line 45: Hoping that nightly code coverage will work soon 😉

PR #1115, opened 2023-08-23
Fixes #1114 Misc issues with package and actions

### LINKED ISSUE #1114: Misc issues with package and actions
* extract installation of R dependencies into separate file **install_dependencies.R** to avoid duplication between appveyor.yml and codcove-GH action
* cleanup/adjust appveyor.yml
  * use **install_dependencies.R**
  * remove coveralls token
  * add GH actions related stuff which is irrelevant for the appveyor build (code covergae, testing under Linux) to the ignore list
* cleanup/adjust code coverage GH action
  * use **install_dependencies.R**
  * fix the issue with the invalid environment expansion in the "show testthat output" step  
grafik
* fix broken "Test under Linux" GH action for Ubuntu

Yuri05:


.github/workflows/codecov-nightly.yml
```diff
@@ -35,11 +31,9 @@ jobs:
           )
         shell: Rscript {0}
       - name: Show testthat output
-        if: always()
```
Yuri05, line 38: I think we don't need such a condition :)

.github/workflows/codecov-nightly.yml
```diff
@@ -35,11 +31,9 @@ jobs:
...
         run: |
-          ## --------------------------------------------------------------------
-          find ${{ runner.temp }}/package -name 'testthat.Rout*' -exec cat '{}' \; || true
-        shell: bash
+          cat ${{ runner.temp }}/package/testthat.Rout*
+        shell: cmd
```
Yuri05, line 36: according to the [documentaion](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsshell) bash should work under windows as well (Git bash is used).
But somehow id did not. So trying cmd instead.

appveyor.yml
```diff
@@ -75,6 +67,11 @@ platform: x64
 skip_commits:
   files:
     - "*.md"
+    - .github/workflows/codecov-nightly.yml
+    - .github/workflows/*.sh
+    - .github/workflows/Dockerfile
+    - .github/workflows/LinuxImages.yml
+    - .github/workflows/testRE.R
```
Yuri05, line 74: adding files which are irrelevant for the appveyour build (tests under Linux; code coverage) to the ignore list

.github/workflows/codecov-nightly.yml
pchelle: Okay, I am wondering if this does not come from the runner.temp ? I got slash/backslash issues testing on my fork within the call of `RUNNER_TEMP` and the error message seemed to indicate a path that misses a slash.

pchelle (approved):

.github/workflows/codecov-nightly.yml
Yuri05: Somehow it comes from the combination of `runner.temp` and Windows.
I think it's a bug of GitHub. 
If it will not work with the cmd as well - I'm out of ideas.

PR #1125, opened 2023-10-12
Fix default angle for y2label

Felixmil:

PR #1126, opened 2023-10-12
Revert "Fix default angle for y2label"

Felixmil:
Reverts Open-Systems-Pharmacology/OSPSuite.ReportingEngine#1125

The change of angle reference for right axis is now handled by tlf


msevestre:
This was a joke before small changes in TLF are always interesting :) closing

Felixmil:
We actually need to revert the value here. Otherwise the y label will be upside down

Issue #1118, opened 2023-09-08
Mass balance plots: configurable splitting


Yuri05:
At the moment, the compartments which are used for the summation of the drug amounts are fixed.
<details><summary><b>Example</b></summary>

![grafik](https://github.com/Open-Systems-Pharmacology/OSPSuite.ReportingEngine/assets/25061876/59110efb-352b-4b61-83ec-a76c6d7fb24c)
</details> 

The user wish to allow the definition of groups of container/molecules which should be summed up and displayed in the mass balance plots.

For each plot, it should be possible to specify all {container, molecule} combinations which should appear in the plot (and give a name to every such combination).

E.g., if in my model there are 2 compounds: **C1** and **C2** and 1 metabolite per compound (**C1-CYP3A4-Metabolite** and **C2-CYP3A4-Metabolite**), it should be possible to define the grouping like this:
* Plot 1
  * **C1** and **C1-CYP3A4-Metabolite** in every plasma/interstitial/intracellular/blood cells compartments - except for brain
  * **C1** and **C1-CYP3A4-Metabolite** in every plasma/interstitial/intracellular/blood cells compartments of the brain
  * **C1** in feces and urine
  * **C1** in the rest of the system (= all containers which are not included into the previous definitions)
* Plot2 (same groping as plot 1, but for **C2** and **C2-CYP3A4-Metabolite**)

Possible specification for such a grouping:
* List of mass balance plot configurations. Each configuration has the properties:
  * Name of the plot (to be displayed in the report)
  * List of subgroupings. Each subgrouping has the properties:
    * Name of the subgrouping (to be displayed in the report)
    * List of paths to include. Each path in the ospsuite-R recognizable format (using the wildcards `*` and `**` ). Must contain at least one entry
    * List of paths to exclude. Each path in the ospsuite-R recognizable format. May be empty.
  * "Rest"-subgroup with the properties:
    * Name to be used in the report
    * List of the **molecule** names 
 
<details><summary><b>Json code for the mass balance settings for the example above</b></summary>

![grafik](https://github.com/Open-Systems-Pharmacology/OSPSuite.ReportingEngine/assets/25061876/73b29876-0f1d-4286-aa4d-2ee4245dd1c8)

```
{
  "Mass balance plots": [
    {
      "Name": "C1 mass balance",
      "Groupings": [
        {
          "Name": "Brainless",
          "Include": [
            "Organism|**|Plasma|C1*",
[...]

PR #1136, opened 2023-11-07
1118 mass balance

pchelle:


inst/extdata/mass-balance-settings.json
```diff
@@ -0,0 +1,100 @@
+{
+    "MassBalancePlots": [
+        {
+            "Name": "Name of mass balance sub-section (optional)",
```
Yuri05, line 4: Is this name used in the report?

inst/extdata/mass-balance-settings.json
```diff
@@ -0,0 +1,100 @@
...
+            ],
+            "Groupings": [
+                {
+                    "Name": "Intracellular",
+                    "Include": [
+                        "Organism|**|Intracellular|Aciclovir*"
+                    ],
+                    "Exclude":[
+                        "Organism|Lumen|*|Aciclovir*"
+                    ]
```
Yuri05, line 16: It is not wrong, however the Exclude section does not make much sense here, because there is no **Intracellular** compartment in the lumen.

inst/extdata/mass-balance-settings.json
```diff
@@ -0,0 +1,100 @@
...
+                },
+                {
+                    "Name": "Plasma & Blood Cells",
+                    "Include": [
+                        "Organism|**|Plasma|Aciclovir*",
+                        "Organism|**|BloodCells|Aciclovir*"
+                    ],
+                    "Exclude":[
+                        "Organism|Lumen|*|Aciclovir*"
+                    ]
```
Yuri05, line 26: Same (no **Plasma** and **BloodCells** compartments in the lumen)

inst/extdata/mass-balance-settings.json
```diff
@@ -0,0 +1,100 @@
...
+                },
+                {
+                    "Name": "Interstitial",
+                    "Include": [
+                        "Organism|**|Interstitial|Aciclovir*"
+                    ],
+                    "Exclude":[
+                        "Organism|Lumen|*|Aciclovir*"
+                    ]
```
Yuri05, line 35: Same (no **Interstitial** compartment in the lumen)

inst/extdata/mass-balance-settings.json
```diff
@@ -0,0 +1,100 @@
...
+            "Name": "Aciclovir Mass Balance focus on Fat",
+            "Molecules": [
+                "Aciclovir"
+            ],
+            "Groupings": [
+                {
+                    "Name": "Fat",
+                    "Include": [
+                        "Organism|Fat|*|Aciclovir",
+                        "Organism|Fat|*|Aciclovir"
```
Yuri05, line 71: why do we have `"Organism|Fat|*|Aciclovir"` 2 times in the list?

inst/extdata/mass-balance-settings.json
```diff
@@ -0,0 +1,100 @@
...
+                    ]
+                },
+                {
+                    "Name": "Plasma & Blood Cells",
+                    "Include": [
+                        "Organism|**|Plasma|Aciclovir",
+                        "Organism|**|BloodCells|Aciclovir"
+                    ],
+                    "Exclude":[
+                        "Organism|Lumen|*|Aciclovir"
```
Yuri05, line 81: Again, Lumen-Exclusion does not have any effect here.
Instead, we should exclude Fat (otherwise it will be represented twice)

inst/extdata/mass-balance-settings.json
```diff
@@ -0,0 +1,100 @@
...
+        {
+            "Name": "Name of mass balance sub-section (optional)",
+            "Molecules": [
+                "Aciclovir"
+            ],
+            "Groupings": [
+                {
+                    "Name": "Intracellular",
+                    "Include": [
+                        "Organism|**|Intracellular|Aciclovir*"
```
Yuri05, line 12: I forgot what we decided: if the molecules list above includes only `Aciclovir`, but I have e.g. `Aciclovir-XY-Metabolite` in my simulation: will `Aciclovir-XY-Metabolite` be included via `"Organism|**|Intracellular|Aciclovir*"` or not?

R/utilities-mass-balance.R
```diff
@@ -8,277 +8,238 @@
...
-  # In the Matlab version, there are multiple groupings performed at that point:
-  # 1) Saliva is excluded from the mass balance calculation
-  # 2) Lumen is a parent compartment that regroups many sub-compartments but Feces
-  # 3) All the rest of the sub-compartments are used normally
 
-  # 1) Exclude Saliva
-  salivaExclusionCondition <- !pathsArray[, "parentCompartmentName"] %in% "Saliva"
-  pathsArray <- pathsArray[salivaExclusionCondition, ]
+    # Get drugMass over time for compoundNames to normalize the mass balance plots
+    applications <- simulation$allApplicationsFor(head(compoundNames, 1))
```
Yuri05, line 51: why `head(compoundNames, 1)` ?? It could retrieve a molecule which is not even being applied.
First we need to build an intersection between `compoundNames `and `allAppliedCompoundNames`
After that, there are 3 cases:
1. The intersection list is empty - Error
2. The intersection list contains exactly 1 molecule - take this molecule for `simulation$allApplicationsFor()`
3. The intersection list contains > 1 molecule: take the **union** of applications for all applied molecules and additionally show a warning (">1 applied molecule in the mass balance grouping, results might be innacurate")

R/utilities-mass-balance.R
```diff
@@ -8,277 +8,238 @@
...
+        dimension = "Time",
+        unit = structureSet$simulationSet$timeUnit
+      ),
+      "Amount" = list(
+        dimension = applications[[1]]$drugMass$dimension,
+        unit = applications[[1]]$drugMass$unit
+      ),
+      "NormalizedAmount" = list(
+        dimension = applications[[1]]$drugMass$dimension,
+        unit = "fraction of drugmass"
```
Yuri05, line 77: dimension should be "Fraction" here and not `drugMass$dimension`, right?

R/utilities-mass-balance.R
```diff
@@ -8,277 +8,238 @@
...
+    massBalanceData$DrugMass <- approx(
+      x = ospsuite::toUnit(
+        "Time",
+        applicationResults$time,
+        structureSet$simulationSet$timeUnit
+      ),
+      y = applicationResults$totalDrugMass,
+      xout = massBalanceData$Time,
+      method = "constant",
+      rule = 2
```
Yuri05, line 92: can you extend the comment explaining what "rule = 2" means?

R/utilities-mass-balance.R
```diff
@@ -357,3 +318,162 @@ getMassBalanceDataMapping <- function(plotType) {
...
+#' @description Get a data.frame of application results corresponding to total drug mass as a function of time.
+#' Accounts for infusion cases in which infusion time is not `NULL` by including intermediate time steps
+#' @param applications list of `Application` objects queried by the method `simulation$allApplicationsFor()`
+#' @param infusionSteps number of time steps to discretize the infusion time
+#' @return A data.frame that includes `time`, `drugMass` and `totalDrugMass` as variables
+#' @import ospsuite
+#' @import ospsuite.utils
+#' @import dplyr
+#' @keywords internal
+getApplicationResults <- function(applications, infusionSteps = 20) {
```
Yuri05, line 332: let's make our life simple and calculate `drugmass(t)=sum_of_drugmass_of_all_applications_starting at Time<=t`
If we start to account for the drugmass which was **effectively** applied (taking the infusion rate into account) - then we would e.g. need to do the same for tablets, which are also dissolved with time - which in general cannot be even estimated apriori without the simulation.

And for infusion, the current implementation is wrong anyways.
I tried it for a single infusion administration in a simulation (infusion time 10 minutes) and got this:
```
> getApplicationResults(applications)
         time  drugMass totalDrugMass
1   0.0000000 0.1560549     0.1560549
2   0.5263158 0.1560549     0.3121099
3   1.0526316 0.1560549     0.4681648
4   1.5789474 0.1560549     0.6242197
5   2.1052632 0.1560549     0.7802747
6   2.6315789 0.1560549     0.9363296
7   3.1578947 0.1560549     1.0923845
8   3.6842105 0.1560549     1.2484395
9   4.2105263 0.1560549     1.4044944
10  4.7368421 0.1560549     1.5605493
11  5.2631579 0.1560549     1.7166042
12  5.7894737 0.1560549     1.8726592
13  6.3157895 0.1560549     2.0287141
14  6.8421053 0.1560549     2.1847690
15  7.3684211 0.1560549     2.3408240
16  7.8947368 0.1560549     2.4968789
17  8.4210526 0.1560549     2.6529338
18  8.9473684 0.1560549     2.8089888
19  9.4736842 0.1560549     2.9650437
20 10.0000000 0.1560549     3.1210986
```

R/utilities-mass-balance.R
```diff
@@ -357,3 +318,162 @@ getMassBalanceDataMapping <- function(plotType) {
...
+        excludedMoleculePaths,
+        sapply(excludedMolecules, function(molecule){molecule$path})
+      )
+    }
+    includedMolecules <- includedMolecules[!includedMoleculePaths %in% excludedMoleculePaths]
+    includedMoleculePaths <- setdiff(includedMoleculePaths, excludedMoleculePaths)
+    
+    if(isEmpty(includedMoleculePaths)) {
+      next
+    }
```
Yuri05, line 404: here we should produce a warning that there is nothing to display in the group (might be a user misconfiguration)

R/utilities-mass-balance.R
```diff
@@ -357,3 +318,162 @@ getMassBalanceDataMapping <- function(plotType) {
...
+#' @param compoundNames Names of simulation molecules
+#' @import ospsuite
+#' @keywords internal
+defaultMassBalanceGroupings <- function(simulationResults, compoundNames) {
+  groupNames <- unique(sapply(
+    simulationResults$allQuantityPaths,
+    function(path) {
+      head(tail(ospsuite::toPathArray(path), 2), 1)
+    }
+  ))
```
Yuri05, line 451: seems wrong to me.
This should return something like {"Interstitial", "Intracellular", "Plasma", ...}, correct? But it returns the name of a compound instead.

```
> unique(sapply(
+     simulationResults$allQuantityPaths,
+     function(path) {
+       head(tail(ospsuite::toPathArray(path), 2), 1)
+     }
+   ))
[1] "Alfentanil"
```

Yuri05 (changes requested):
s. comments

inst/extdata/mass-balance-settings.json
pchelle: If not empty, yes, the name will be displayed in the report as a sub section

inst/extdata/mass-balance-settings.json
pchelle: Then, I'll remove

inst/extdata/mass-balance-settings.json
pchelle: Copy/paste typo or copilot AI auto fill. I'll remove

inst/extdata/mass-balance-settings.json
pchelle: If Fat was included in a previous field, it won't be included here.
In issue #1118, we suggested the optional field `ExcludePreviousGroupings` which is true by default.

I'll remove all the redundant Lumen exclusions

inst/extdata/mass-balance-settings.json
pchelle: Yes, I stated it in the vignette


Note that inclusion and exclusion criteria for molecule paths leverage [wild card symbols (`*`)](https://www.open-systems-pharmacology.org/OSPSuite-R/articles/load-get.html#accessing-entities-of-the-model-and-their-properties---the-path-concept).
...
Another use case is when a metabolite is defined using the molecule name as prefix.
Then, the metabolite can be included in the grouping.

```json
{
  "Name": "Plasma for Aciclovir and metabolites", 
  "Include": ["Organism|**|Plasma|Aciclovir*"]
}
```

R/utilities-mass-balance.R
pchelle: I agree with 1. and 2.

Regarding 3., there is currently an issue with the `{ospsuite}` package: if `compoundNames` includes a molecule and its metabolites. We get a duplication of the applications because the method retrieves the same applications for the molecule and the metabolites.

Yuri05: ok, let's handle 3. as 1. for now: error
The use cases where 3. makes sense are rare anyways.

pchelle: For this reason, I just kept the first compound name thinking that the other will likely be the metabolites and get the same applications.
(However since applications are lists of R6/.NET objects, assessing if they are the same might be tricky)

R/utilities-mass-balance.R
pchelle: Here, I am simply leveraging the smart mapping from tlf that will display `dimension [unit]` in the plot.
This way we get `Amount [fraction of drugmass]` displayed in the plot axis.
Let me know if you wish to display the normalized plot differently

R/utilities-mass-balance.R
pchelle: Yes, I'll include that in the code as well.
To keep track, I'll also write here: the argument `rule` defines what happens in case you extrapolate `yout` outside of the smallest/biggest `x` value. `rule=1` will predict `NA` and `rule=2` will predict `y` value of smallest/biggest `x` value.

inst/extdata/mass-balance-settings.json
Yuri05: What I meant: the molecules list in the example contains `Aciclovir `only:
```
            "Molecules": [
                "Aciclovir"
            ],
```
Now if `"Include": ["Organism|**|Plasma|Aciclovir*"]` retrieves both `Aciclovir` and `Aciclovir-XY-Metabolite`: will the metabolite be included or not?

R/utilities-mass-balance.R
pchelle: There should be a step before that only include molecule paths in the simulation results.
In the matlab code, the compartment to display in such molecule paths were compartment of the before the last element of the path array. 
Is it not actually the case ?

R/utilities-mass-balance.R
Yuri05: I would not rely on this and filter out only those molecules which are applied. 
And then check if the remaining list contains exactly 1 molecule or not.

R/utilities-mass-balance.R
Yuri05: ok, forget it, you are right.
In my test example I had a concentration instead of amount, that's why the returned grouping was wrong

R/utilities-mass-balance.R
Yuri05: ok, let's keep for now like this :)

inst/extdata/mass-balance-settings.json
pchelle: Since only aciclovir was in molecules, no, they should not be retreived from the search.
(because the interesection between all the molecule paths from the compound in Molecules field and the Include field is used).

Yuri05: ok. it would be good to select another example for vignette, which incorporates metabolites as well - but let's keep Aciclovir for now AS IS.

R/utilities-mass-balance.R
```diff
@@ -357,3 +318,162 @@ getMassBalanceDataMapping <- function(plotType) {
...
+  defaultGroupings <- c(
+    lapply(
+      groupNames,
+      function(groupName) {
+        list(
+          Name = paste(paste(compoundNames, collapse = ", "), groupName, sep = " - "),
+          Include = paste0("Organism|**|", groupName, "|", compoundNames),
+          # 1) Exclude saliva and 2) Exclude Lumen
+          Exclude = c(
+            paste0("Organism|Saliva|**|", compoundNames),
```
Yuri05, line 462: Saliva exclusion does not make sense here.
At the moment it's excluded everywhere
grafik

Yuri05: Instead, `groupNames `must be reduced.
At the moment we get the following group names for a standard simulation:
grafik
Highlighted group names should be removed (Saliva+SalivaryGland+Lumen_segements)

Actually, it's easier to define the group names of interest manually, without asking the model:
* "Plasma"
* "BloodCells"
* "Interstitial"
* "Intracellular"
* "Endosome"
* "Gallbladder"
* "Urine"
* "Feces"

Yuri05: Finally: do we have the "Rest" grouping in the default grouping?
If not: it should be added.
And here: exclude Saliva :)

R/utilities-mass-balance.R
```diff
@@ -357,3 +324,153 @@ getMassBalanceDataMapping <- function(plotType) {
...
+    list(list(
+      Name = paste("Rest of", paste(compoundNames, collapse = ", ")),
+      # Since default option excludes previously included paths,
+      # we can include all the paths but saliva here
+      Include = paste0("Organism|**|", compoundNames),
+      # Exclude saliva from rest
+      Exclude = paste0("Organism|Saliva|**|", compoundNames)
+    ))
+  )
+  return(defaultGroupings)
```
Yuri05, line 475: almost there. The "Lumen" grouping is missing now (Lumen except Feces):
```
Include = paste0("Organism|Lumen|*|", compoundNames),
Exclude = paste0("Organism|Lumen|Feces|", compoundNames)
```

R/utilities-mass-balance.R
```diff
@@ -8,277 +8,244 @@
...
+    compoundNames <- as.character(plotSettings$Molecules)
+    # Sub section using settings name
+    if(!isEmpty(plotSettings$Name)){
+      sectionId <- defaultFileNames$resultID(length(massBalanceResults) + 1, "mass_balance")
+      massBalanceResults[[sectionId]] <- saveTaskResults(
+        id = sectionId,
+        textChunk = paste("###", plotSettings$Name, anchor(sectionId)),
+        includeTextChunk = TRUE
       )
+    }
```
Yuri05: what happens if `plotSettings$Name` is empty? 
Somehow I cannot see the else-part (where the **default** subchapter caption is written)

pchelle: Currently, there is no sub section title added if `plotSettings$Name` is empty.

inst/extdata/mass-balance-settings.json
Yuri05: created an issue for that: https://github.com/Open-Systems-Pharmacology/OSPSuite.ReportingEngine/issues/1148

Yuri05 (approved):

Yuri05:
@pchelle 
after the PR was merged 1 hour ago, I still don't see the mass balance article on the website (dev). Any idea why?
grafik

Issue #1015, opened 2023-03-20
lintr warnings


Yuri05:
`lintr::lint_dir(linters=linters_with_defaults(object_name_linter=NULL, line_length_linter=NULL, object_length_linter=NULL, brace_linter=NULL, paren_body_linter=NULL, trailing_whitespace_linter=NULL, trailing_blank_lines_linter=NULL, spaces_left_parentheses_linter=NULL, infix_spaces_linter=NULL))`

![grafik](https://user-images.githubusercontent.com/25061876/226377968-b2db58c4-2942-4d8c-a9ca-f4043d8df2ef.png)
![grafik](https://user-images.githubusercontent.com/25061876/226378099-f86d83e4-4e41-4e91-92de-a756c93178a9.png)
![grafik](https://user-images.githubusercontent.com/25061876/226378237-e770bf42-0823-41a1-8967-99ca837a64db.png)

PR #1158, opened 2023-12-14
Set up lintr github action

pchelle:


.github/workflows/run_lintr.R
```diff
@@ -0,0 +1,13 @@
+lintr::lint_package(
+  linters=lintr::linters_with_defaults(
+    object_name_linter=lintr::object_name_linter(styles=c("camelCase","CamelCase","symbols")),
```
Yuri05, line 3: I th

pchelle:
@Yuri05 
I have set up `lintr` to check the package as a github action.
Hopefully, it will be useful to track down and remove naming issues and bring down the code complexity in future PRs.
Currently, I only fixed the simple suggestions from #1015 

So far, it runs `lint_package` which also looks at the vignettes and tests.
Let me know if they should be removed from the check.
We may also want to add some tuning at some point for `object_name_linter` to better account for class names using the CamelCase naming as well as the use of parameters that are from other packages such as `na.rm`.

Yuri05 (approved):

Yuri05:
@pchelle Looks very good for me! 
Questions/remarks that I have at the moment:
* is it possible to configure object_name_linter in the way that different rules apply for variables and classes/functions? Because for the variables we use only camelCase but not CamelCase
* somehow it did not work out with the badge:
grafik
* Annotations (lint warnings) are apparently limited to 10, while when looking into the job description you will see much more linter warnings. We could consider something like this: https://github.com/orgs/community/discussions/26680#discussioncomment-3252838

PR #1167, opened 2024-01-05
Fixes #1016

### LINKED ISSUE #1016: Minor issue: SVG plots are not properly rendered in the word report
Example s. https://github.com/Yuri05/Digoxin-Model/tree/review/Evaluation/report_svg 
MArkdown report looks good, word report has just empty spaces instead of images

Yuri05:
No functional changes, just an update of installation instructions.

It turned out that for correct conversion of reports with SVG images from .MD to .docx via Pandoc one extra command line utility is required: **rsvg-convert**.
Under Linux it's part of librsvg package, under Windows it can be installed as a standalone command line tool.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ The **OSPSuite.ReportingEngine** package is compatible with version 3.6.x **AND*
 
 [OPTIONAL] Install **rsvg-convert** (required by Pandoc for conversion of images in SVG format)
 
-* [Installer (Windows)]([Releases · miyako/console-rsvg-convert (github.com)](https://github.com/miyako/console-rsvg-convert/releases))
+* [Installer (Windows)](https://github.com/miyako/console-rsvg-convert/releases)
   * The installation folder must be added to the system path.
 * For Linux, **librsvg** package must be installed (package name depends on distribution, e.g. **librsvg2-bin** for Ubuntu).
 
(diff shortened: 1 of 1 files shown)


pchelle:
> Pandoc one extra command line utility is required: **rsvg-convert**.

Does it mean I need to update the config file to include this extra if svg are in use ?

Yuri05:
> Nice ! Have you tried the conversion ?

yes, seems to work so far :)

Yuri05:
> > Pandoc one extra command line utility is required: **rsvg-convert**.
> 
> Does it mean I need to update the config file to include this extra if svg are in use ?

which config file?

pchelle:
> > > Pandoc one extra command line utility is required: **rsvg-convert**.
> > 
> > 
> > Does it mean I need to update the config file to include this extra if svg are in use ?
> 
> which config file?

https://github.com/Open-Systems-Pharmacology/OSPSuite.ReportingEngine/blob/19275741057fa88bea406e69c3af340a77bb70da/R/utilities-writing-report.R#L313-L334

Yuri05:
no, nothing needs to be updated there. 
pandoc calls rsvg-convert automatically when it finds an svg image.
If not found, pandoc produce warnings like this:
`[WARNING] Could not convert image absorption_Raltegravir.svg: "check that rsvg-convert is in path.\nrsvg-convert: createProcess: does not exist (No such file or directory)"`

However RE could check if the utility is correctly installed, e.g.
* If RE image format is svg **and** word conversion is required:
  * execute `ret = system("rsvg-convert --version")`
  * if return value equals 127: rsvg-convert is not found on the system. Show an explicit warning

Issue #840, opened 2022-06-28
Create unit tests for ddi ratio plots


pchelle:
DDI Ratio plots should test at least

- [x] subunits are alphabetically ordered
- [x] subunits are included in TOC
- [x] artifacts are output correctly
- [x] multiple observed dataset formats are handled
- [x] NA values are handled

pchelle:
The 2 last items are tested in `test-observed-data`

PR #1175, opened 2024-01-18
Draft DDI ratio plots unit tests for #840

pchelle:
snapshots sounds the best way to test that artifacts are correctly output, however this may lead to failing test every time the report will change something in the way it writes the report


pchelle:
Interestingly, the DDI qualification workflow internally creates user defined PK parameters.
These PK parameters are not removed when reloading the `{ospsuite}` or `{ospsuite.reportingengine}` library, thus I had to explicitely call `removeAllUserDefinedPKParameters()` in the tests.

@Yuri05 
Would using `.onLoad` be relevant to reset `{ospsuite.reportingengine}` dependencies and clearing cache ?

Yuri05:
@pchelle Is the PR ready to merge or draft?

Yuri05:
>Would using .onLoad be relevant to reset {ospsuite.reportingengine} dependencies and clearing cache ?

I would not do this, might lead to unwanted side effects (e.g. if interactively working with ospsuite-R and adding some user defined pk parameters etc. - I would not want that loading RE resets them)

pchelle:
> @pchelle Is the PR ready to merge or draft?

Ready to review/merge.
The title has "draft" because this does not fix the issue (there are still a few items on the task list that are not tested and that I am still wondering the best way to test)

PR #1178, opened 2024-01-30
Fixes #1177 automate website building and deployment

### LINKED ISSUE #1177: Automate website build
To automate the website build, the github action below may be used in addition to some modifications.

- Creation of the `gh-pages` branch whose updates will only be affected and set source website from this branch
- `webshot::install_phantomjs()` needs to be replaced by `webshot::install_phantomjs(force=TRUE)` to prevent crash when already installed and the function is called.

```yaml
name: pkgdown-dev
on:
  push:
    
jobs:
  pkgdown-dev:
    runs-on: windows-latest
    env:
      GITHUB_PAT: ${{ secrets.GITHUB_TOKEN }}
    permissions:
      contents: write
    steps:
      - name: Checkout
        uses: actions/checkout@v3
      - name: Setup
        uses: r-lib/actions/setup-r@v2
        with:
          use-public-rspm: true
      - name: Setup Pandoc for word conversion
        uses: r-lib/actions/setup-pandoc@v2
      - name: Install ospsuite ecosystem and cran packages
        run: |
          Rscript .github/workflows/install_dependencies.R
      - name: Build dev site
        run: |
          pkgdown::build_site(install=TRUE)
        shell: Rscript {0}
     - name: Deploy to GitHub pages
        uses: JamesIves/github-pages-deploy-action@v4.4.1
        with:
          clean: false
          branch: gh-pages
          folder: docs
```

For release site, it is possible to trigger only when `main` is merged by using:

```yaml
on:
  pull_request:
    branches:
      - main
    types:
      - closed
```

and, in the jobs, include the following in case the PR is closed without merging

```yaml
if: github.event.pull_request.merged == true)
```

pchelle:
A few notes regarding some updated files:
<details><summary>_pkgdown</summary>
Note that the _pkgdown file was in .github to make it easier to find, however it was not saved as yml there to prevent github assuming it is an action.
</details>
<details><summary>DESCRIPTION</summary>
Since pkgdown takes the version from DESCRIPTION as is, it as updated there to 2.2.
If we don't feel comfortable with that we can do the same tweak as the `_pkgdown` file: storing in a different folder and overwriting the file for the sake of the website built only within the scope of the github action.
</details>
<details><summary>vignette-helpers</summary>
When running the github action, webshot::install_phantomjs crashed because it was not able to find the location of phantom JS between vignettes and crashed now and then. Forcing the installation for each vignette ensures the location is found and udpated. 
</details>


pchelle:
@Yuri05 
It seems the workflow cannot update the `gh-pages` branch because it does not have the [write permission](https://github.com/Open-Systems-Pharmacology/OSPSuite.ReportingEngine/actions/runs/7710519970/job/21013993908)

I do not think I have the credentials to update such permissions (and test what is advised here: https://docs.github.com/en/actions/security-guides/automatic-token-authentication#modifying-the-permissions-for-the-github_token)

.github/workflows/update-dev-website.yml
```diff
@@ -0,0 +1,60 @@
+name: Update dev website
+on:
+  [push, pull_request]
```
Yuri05, line 3: The workflow must be restricted to **develop** branch only (similar like the release workflow is restricted to the main branch)

.github/workflows/update-dev-website.yml
```diff
@@ -0,0 +1,60 @@
...
+          path: docs
+
+  deploy:
+    concurrency: ci-${{ github.ref }}
+    # The second job must depend on the first one to complete before running and uses ubuntu-latest instead of windows.
+    needs: [build] 
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
```
Yuri05, line 45: why using checkout@**v3** and not checkout@**v4** ?

Yuri05: Actually why do we need this second checkout at all?

DESCRIPTION
```diff
@@ -1,7 +1,7 @@
 Type: Package
 Package: ospsuite.reportingengine
 Title: R package to create reports for the Open Systems Pharmacology models
-Version: 2.2.0
+Version: 2.2
```
Yuri05, line 4: The DESCRIPTION file must remain unchanged. 
The **GitHub-Action** should remove **.0** in the file before calling `build_site()`

.github/workflows/_pkgdown
Yuri05, line 1: I would not create a copy of this file (2 files will not be properly synchronised one day for sure).
Either the GH-action should copy the file and replace **mode: devel** with **mode: release**

Or we completely remove the mode statement from the file (or set it to auto - don't remember what's the proper way)
Instead **build_site()** should be called with **devel = FALSE** for the release build and with **devel = TRUE** for the dev build.

Yuri05: ok, so what would work:
* define environment variable (e.g. **PKGDOWN_MODE**)
* in the workflow, set **PKGDOWN_MODE** either to **devel** or to **release**
* in **_pkgdown.yaml** get the mode from this environment variable:
```
development:
  mode: !expr Sys.getenv('PKGDOWN_MODE')
```
* before calling **build_site()**: make sure to set the R option for yaml for evaluating expressions:
`options(yaml.eval.expr=TRUE)`

Perfectly works for me when doing this in R :)

Yuri05:
> @Yuri05 It seems the workflow cannot update the `gh-pages` branch because it does not have the [write permission](https://github.com/Open-Systems-Pharmacology/OSPSuite.ReportingEngine/actions/runs/7710519970/job/21013993908)
> 
> I do not think I have the credentials to update such permissions (and test what is advised here: https://docs.github.com/en/actions/security-guides/automatic-token-authentication#modifying-the-permissions-for-the-github_token)

I think you need to pass the GH-token to the deploy action

E.g. something like
```
...
        with:
          token: ${{ secrets.GITHUB_TOKEN }}
          folder: docs
          clean: false
          branch: gh-pages
          target-folder: docs
...
```

Yuri05:
@pchelle 
I guess this is the reason for the missing permissions:
https://docs.github.com/en/actions/security-guides/automatic-token-authentication
grafik

So I would suggest to keep only `on push` trigger and remove the `on pull_request` trigger. Then we merge once and check if it works when triggered by `on push`

PR #1187, opened 2024-02-29
1183 reset settings

pchelle:


pchelle:
@Yuri05 
appveyor failed because TLF was not updated yet (`resetTLFSettingsToDefault()` was not available).
Can we re-run appveyor ?

Yuri05:
> @Yuri05 appveyor failed because TLF was not updated yet (`resetTLFSettingsToDefault()` was not available). Can we re-run appveyor ?

Done, succeedeed on rebuild :)

R/reportingengine-env.R
```diff
@@ -352,97 +228,21 @@ getStatisticsFromType <- function(statisticsType) {
...
-  validateIsInteger(n)
-  reEnv$defaultMCRepetitions <- n
-  return(invisible())
-}
+# Save a sync version of re-env as an .RData file
+newEnv <- reEnv
+save(
+  "newEnv",
+  file = file.path(system.file(package = "ospsuite.reportingengine"), "extdata", "re-env.RData")
+)
```
Yuri05, line 247: When is this called?
We should assume that the package installation location is read-only and don't write anything there during runtime.

pchelle: It is called when I update the documentation of the package: `devtools::document()` and not exported/exposed to the user.

Yuri05 (approved):

Yuri05:
@pchelle with this PR and the latest tlf I can retry to generate the Quali report, correct?

PR #1222, opened 2024-05-16
Fixes #1209 allow different outputs for pop PK parameter plots

### LINKED ISSUE #1209: Crash of PlotPKParameter because of different outputs in simulations
In my example I have a workflow with simulations with one out put and simulations with two outputs.
That crashes in the function validateSameOutputsBetweenSets in the line 
"if (all(pkParametersTable$path == pkParametersTableRef$path)) "  as the data.frames have a different size, "==" does not work. The error message is "longer object length is not a multiple of shorter object length"

But why do we have this condition on same outputs between sets.
We produce one plot per simualtion and output.  I do not see a reason why we can not mix simulations with different output numbers.
Also the number of PK Parameter per output shoudn't be an issue. Why can we not have some plots with e.g. 6 simulations on the x-axis, and some with e.g. 4?

pchelle:


Yuri05:
@pchelle Do we have any test for the "different outputs" scenario? If not: please add one.

Yuri05 (approved):

PR #1232, opened 2024-06-19
1230 time offset

pchelle:


R/utilities-simulation-set.R
```diff
@@ -199,10 +199,8 @@ getSimulationTimeRanges <- function(simulation, path, simulationSet) {
   logDebug(messages$numberOfApplications(length(applications), path, simulation$name))
   logDebug(messages$timeRangesForSimulation(paste0(simulationRanges, collapse = "', '"), simulation$name))
 
-  # Define ranges for output
-  # Depending on expected behaviour of settings$applicationRange
-  # It would be possible to set these values
-  timeRanges$total$values <- c(min(simulationRanges), max(simulationRanges))
+  # Define ranges for output based on time offset and simulation
+  timeRanges$total$values <- c(simulationSet$timeOffset, max(simulationRanges))
```
Yuri05, line 203: I think this is wrong. What if the first application is given not at t=0?
Should be something like `c(min(simulationRanges, simulationSet$timeOffset), max(simulationRanges))`

R/utilities-simulation-set.R
Yuri05, line 1: btw: do we have somewhere a check that the time offset is strictly < simulation end time?

pchelle: When creating a simulation set, the only check is that `timeOffset` is numeric, we can include `>=0` and `<simulation end time`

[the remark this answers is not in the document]
pchelle: We currently do not update the output schema. So, the output time points won't necessarily include the offset

R/utilities-simulation-set.R
pchelle: Then, I have a very naive question.
Was the goal of the offset only to start at the time of first application if the application started >0 ? 
If so, is there actually a need for the offset ?

R/utilities-simulation-set.R
Yuri05: let's do

R/utilities-simulation-set.R
KatrinCoboeken: Yes, we could produce a warnig, I also could not think of a use case where time offset is not equal to an application start time.

R/utilities-simulation-set.R
KatrinCoboeken: &#x1F44D;

Yuri05:
@pchelle Btw, it is indeed possible to add single time points to the simulation output schema via https://www.open-systems-pharmacology.org/OSPSuite-R/dev/reference/OutputSchema.html#method-addtimepoints-
However it's not quite clear for me, if the added values are automatically assumed as the internal time unit ([min]) or units can be passed additionally (just created issues for this in ospsuite-r)

R/simulation-set.R
```diff
@@ -55,39 +55,45 @@ SimulationSet <- R6::R6Class(
...
       validateFileExists(simulationFile)
       simulation <- ospsuite::loadSimulation(simulationFile, addToCache = FALSE)
-
+      # Checks requiring simulation info
+      endTime <- ospsuite::toUnit(
+        quantityOrDimension = "Time",
+        values = simulation$outputSchema$endTime,
+        targetUnit = timeUnit
+      )
+      validateVectorRange(timeOffset, type = "numeric", valueRange = c(0, endTime))
```
Yuri05, line 79: Question: if the `minimumSimulationEndTime` is passed: is the simulation end time adjusted **before** or **after** this check? Because if it's done after the range bound should be `max(simulation$outputSchema$endTime, minimumSimulationEndTime)`

R/utilities-simulation-set.R
Yuri05, line 1: Question not related to the current PR: in the lines 132 and 136, `minimumSimulationEndTime` is assumed to be in minutes
https://github.com/Open-Systems-Pharmacology/OSPSuite.ReportingEngine/blob/d8efa073e8b502c4aaee9a0c25f5dbc62735327c/R/utilities-simulation-set.R#L132-L136

However when passed to the simulation set constructor, `minimumSimulationEndTime` is assumed to be in the user defined unit `timeUnit` (which is also [h] as per default).
https://github.com/Open-Systems-Pharmacology/OSPSuite.ReportingEngine/blob/d8efa073e8b502c4aaee9a0c25f5dbc62735327c/R/simulation-set.R#L16-L28

As long as there is no unit conversion when setting `minimumSimulationEndTime` it seems to be a bug?
@pchelle

R/simulation-set.R
pchelle: Oups, before.
Also, I noticed I forced `minimumSimulationEndTime` to be within 0 and `outputSchema$endTime`...
What would be the expected range for this argument actually ? `>0`, `[0, endTime]`, or `[endTime, Inf]`

R/utilities-simulation-set.R
pchelle: I agree. `minimumSimulationEndTime ` should be defined by user in `timeUnit`, and be converted in `baseUnit` in this function.
I created issue #1238 to fix this

R/simulation-set.R
Yuri05: I think it should be `>0`
Also if `minimumSimulationEndTime <= endTime` I actually would not change the simulation end time.

Yuri05 (approved):

PR #1254, opened 2024-07-22
Fixes #1252 use <br> tag instead of \n for linebreak

### LINKED ISSUE #1252: Line break in residual plots
In the resulting reports of the reporting engine, the residuals plots for single simulation sets have a line break in the ylabel, the plots in the residuals acrross all simulatios don't. I would consistently add a linebreak between "Residuals" and "log(simulated)-log(observed)" or "simulated-observed"

e.g. in this test repost:
https://github.com/Yuri05/TestReport/blob/main/Report_20_MeanModel_Raltegravir/Report.md#time-profiles-raltegravir_10_mg__lactose_formulation_
 Figure 5-5 vs. Figure 5-97
 
For the x-axis there is no line break at all: Figure 5-7 and  Figure 5-99 - I would also use the same line break here

pchelle:


R/gof-plot-task.R
```diff
@@ -256,9 +256,9 @@ GofPlotTask <- R6::R6Class(
 
       # Plot the residuals across the simulations by output path
       for (output in allOutputs) {
-        Legend <- "Residuals\nlog(observed)-log(simulated)"
+        Legend <- "Residuals<br>log(observed)-log(simulated)"
```
Yuri05, line 259: let's move all these captions to `captions.R` and reuse then!

Yuri05 (approved):

PR #1272, opened 2024-09-03
Fixes #1271 remove blanks before anchors in titles

### LINKED ISSUE #1271: Remove the blank for added subchapter anchors
When adding anchors to subchapters: remove the blank between the subchapter title and the anchor. E.g. 
`# 1 Introduction<a id="1"></a>`
instead of 
`# 1 Introduction <a id="1"></a>`

Otherwise when users try to reference a chapter by their title, those links easily become invalid
(e.g. in the example above user would use `#1-introduction` as reference - but in fact the correct reference would have a "-" at the end: `#1-introduction-` - which is not transparent.

pchelle:


Yuri05 (approved):

Yuri05:
@pchelle 8 tests failed...

Yuri05 (changes requested):
fix failinbg tests

pchelle:
> fix failing tests

Fixed now :)

Yuri05 (approved):

PR #1297, opened 2025-03-05
Fixes #1296 switch from appveyor to github workflow

pchelle:

PR #1300, opened 2025-03-06
Fixes #1299 consistent exported image size

### LINKED ISSUE #1299: Consistent exported image size
See OSP-PBPK-Model-Library issue [#125](https://github.com/Open-Systems-Pharmacology/OSP-PBPK-Model-Library/issues/125) for more details

pchelle:
PR #1297 may be needed first


pchelle:
Here is the result from latest test:

| Local | Github Action |
|---|---|
| 3_time_profile_plot_7E3_7E3_WildTypeMouse | 3_time_profile_plot_7E3_7E3_WildTypeMouse |

We also previously defined a __golden number__ `reEnv$fontScaleFactor` that aimed at resizing the final plot to get a nice rendering.
The current value is 2 creating default plots of 1000x800 pixels.
Accounting for the previous conversion factor and rendering, I think a value of 4 or 5 should render nicely. I'll attached below once I have re-run with the new number

pchelle:
Here are the same plot when tweaking the golden number `reEnv$fontScaleFactor` (and applying it only to plot dimensions).
Which size seems more appropriate ?

| `fontScaleFactor=3` | `fontScaleFactor=3.5` | `fontScaleFactor=4` |
|---|---|---|
| 3_time_profile_plot_7E3_7E3_WildTypeMouse | 3_time_profile_plot_7E3_7E3_WildTypeMouse | 3_time_profile_plot_7E3_7E3_WildTypeMouse |

Yuri05:
> Here are the same plot when tweaking the golden number `reEnv$fontScaleFactor` (and applying it only to plot dimensions).
> Which size seems more appropriate ?

Hard choice... can you also add a picture with factor 5 into the comparison table above?

pchelle:
> > Here are the same plot when tweaking the golden number `reEnv$fontScaleFactor` (and applying it only to plot dimensions).
> > Which size seems more appropriate ?
> 
> Hard choice... can you also add a picture with factor 5 into the comparison table above?

Updated above

Yuri05:
> Which size seems more appropriate ?

I think factor 5 is the best... @AnnikaRPS @KatrinCoboeken Do you have any opinion on this?
I have also created a dummy report with those 4 pictures: https://github.com/Yuri05/Test20/blob/main/test.md 

@pchelle We need to improve the legends entry splitting in the plots... in the last 2 plots (factor 4, factor 5) the first legend entry fits into the image width without splitting

pchelle:
> @pchelle We need to improve the legends entry splitting in the plots... in the last 2 plots (factor 4, factor 5) the first legend entry fits into the image width without splitting

I totally agree and I need to investigate it further.
I did a dummy test of our splitting algorithm with the expected sizes as input and was not supposed to split for scale factors of 4 and 5 as we can see on the plots.

Yuri05 (approved):

PR #1302, opened 2025-03-14
1301 permission

pchelle:


.github/workflows/main-workflow.yaml
```diff
@@ -5,7 +5,7 @@ on:
     branches: [main]
   pull_request:
 
-permissions: write-all
+permissions: write
```
Yuri05, line 8: wrong syntax. must be
```
permissions:
  contents: write
```

pchelle: Oups, my bad.

PR #1311, opened 2025-03-20
1304 workflow

pchelle:


.github/workflows/main-workflow.yaml
```diff
@@ -5,9 +5,24 @@ on:
...
+  issues: read
+  packages: read
+  pages: read
+  pull-requests: read
+  repository-projects: read
+  statuses: read
+  security-events: read
+  # read does not seem to exist for id-token
+  id-token: write
+  models: read
```
pchelle, line 24: @Yuri05 the main workflow did not run initially
The reason was the the `R-CMD-Check` and `coverage` actions require a `read-all` 
and `bump` and `pkgdown` require `read` for all except `content` that requires `write`

In the end, I had to define them all one by one because of that.
Additionally, the `id-token` permission does not accept a `read` (either `write` or `none` but the actions required `read` at least)
But `id-token` does not have a `read`

.github/workflows/lint.yaml
```diff
@@ -1,32 +1,66 @@
...
+        # Use the following recommendations
+        # https://docs.github.com/fr/rest/checks/runs?apiVersion=2022-11-28#update-a-check-run
+        run: |
+          curl -L \
+            -X PATCH \
+            -H "Accept: application/vnd.github+json" \
+            -H "Authorization: Bearer ${{ secrets.GITHUB_TOKEN }}" \
+            -H "X-GitHub-Api-Version: 2022-11-28" \
+            https://api.github.com/repos${{ github.owner }}/${{ github.repository }}/check-runs/${{ steps.job-id.outputs.job-id }} \
+            -d '{"name":"lint","status":"completed","conclusion": "neutral"}'
```
pchelle, line 65: @Yuri05 
I wanted to implement a soft check for the lint action displaying something different than &check; or &cross; if the code throws warnings from the lint check (called `neutral` in the documentation).
This way, the list of checks in the PR would display neutral 🔲 and the readme badge would show a failing status.

However, to update the conclusion of the check, the `GITHUB_TOKEN` permissions for `checks` needs to be `write`.
So it does not currently work here but it works in my repo: https://github.com/pchelle/OSPSuite.ReportingEngine/actions/runs/13988905247/job/39168176828

and here would be the badge: [lint.yaml](https://github.com/pchelle/OSPSuite.ReportingEngine/actions/workflows/lint.yaml)

PR #1328, opened 2025-04-17
Fix trim file name

### LINKED ISSUE #1327: Fix `trimFileName`
It looks like 4.5.0 changed the way unnamed arguments are used. In the code above, I think there is a bug because this 

https://github.com/Open-Systems-Pharmacology/OSPSuite.ReportingEngine/blob/344b0876623f2dd43295f6fd3ae60d7cbe8997e7/R/utils.R#L87

should not be there at all. In R 4.4 this was ignored (??), but in 4.5.0  the first unnamed argument is assigned to the first unused argument of the function, or so... So ` "[.].*$"` is assigned to the argument `ignore.case`, which is expected to be a boolean.

_Originally posted by @PavelBal in https://github.com/Open-Systems-Pharmacology/Amikacin-Model/issues/9#issuecomment-2812449014_

PavelBal:
Fixes #1327


inst/extdata/re-env.RData
Yuri05, line 1: What was changed here?

PavelBal: This file again... I have no idea what this is and especially where it get's generated? @pchelle how is it suppose to work? Every time I commit something this file changes...

pchelle: Default settings are accessed by reporting engine through the `reEnv` environment object.
> for instance, `reEnv$defaultPlotFormat` defines the default height, width, dpi and extension of exported figures

Since it is an environment object (that include R6 objects for plot themes, logging system, etc), the simplest way I have found to store it, is through this `.RData` file.

When running `devtools::document()`, a new `reEnv` object is created and stored into the `.RData` file. It seems that each environment is unique and leads to a different file even if the settings are the same.

pchelle: A potential solution would be to prevent the automated update of the file.
However, if someone change the default settings, they have to update the `.RData` file.

https://github.com/Open-Systems-Pharmacology/OSPSuite.ReportingEngine/blob/344b0876623f2dd43295f6fd3ae60d7cbe8997e7/R/reportingengine-env.R#L254-L260

pchelle: @PavelBal could you revert the commit on the `re-env.RData` file ?

To prevent further issue, I made PR #1330 that will prevent this automated update

PavelBal: @pchelle done

Yuri05 (approved):

PavelBal:
lol I cannot merge.

Yuri05:
> lol I cannot merge.

Will merge after approval by @pchelle :)

pchelle (approved):

